Issue #27, opened 2015-11-05
Persistent blob urls


petkaantonov:
*Update: See https://gist.github.com/petkaantonov/37a9c64ea2bbd1aa3ca5* 

<del>Currently blob urls don't persist between browser restarts / page refreshes etc so the user has to add same files to the app over and over again everytime the app is started.</del>

<del>For certain kinds of apps it is necessary to be able to create a persistent blob url. For instance a media player app that has access to the media files the user has added to the app so they don't have to keep readding the files (possibly thousands) every time the app is started. Duplicating copies of the files in FileSystem or IndexedDB is out of the question as this duplicates possibly gigabytes of data which takes a long time, wastes a lot of storage and hits limits quickly.</del>

<del>**Reading**</del>

<del>I suggest a method like `URL.createPersistentObjectUrl` is added. The url is valid as long as the original file exists in exactly same location with exactly same attributes and contents as they were when the URL was created. When the file is modified, moved or altered in any way, the url should become invalid (perhaps `URL.isPersistentUrlValid` can be used to check it?). This can be probably done just by checking the last modified attribute.</del>

<del>**Writing**</del>

<del>One could also be allowed to write to a valid persistent object urls by sending binary data as a POST request to the url. However, *every* time such a POST request is made, the browser should prompt the user for confirmation that "the app wants to modify the flie at [path], is this ok?". Also such requests should probably only be allowed to made in response to user event like click. After successful write, the browser should update the persistent object url to match the new file attributes and contents that were updated.</del>

<del>Security considerations</del>

<del>As far as I can see, this should not give the app access to any more information than existing APIs already give.</del>

annevk:
It seems better to just store the files in FileSystem or IndexedDB and let the browser worry about deduplication. That's what we recommend elsewhere.

petkaantonov:
Deduplication is absolutely required, not an optional optimization that browser could do if it wants. And it doesn't address the writing part, which is better experience than downloading, browsing and clicking the file to replace.

petkaantonov:
[A chromium developer said](https://code.google.com/p/chromium/issues/detail?id=108012#c69) that what is proposed here is fundamentally impossible with IndexedDB.

I also don't see anything about FileSystem API that deals with references to actual files in the actual file system like `File` objects behave currently. What is proposed is `File` objects current behavior retained across refreshes and browser restarts (protected with e.g. ctime attribute), this is nothing like the FileSystem API.

petkaantonov:
[...]

PR #31, opened 2016-02-02
Make the Blob URL creating/revoking methods not be exposed on the ser…

jungkees:
…vice worker global.

There has been a discussion around what the lifetime of a blob URL created in a service worker would be: https://github.com/slightlyoff/ServiceWorker/issues/688

During the f2f discussion last week, the service worker folks reached a consensus that the methods creating/revoking a blob URL should not be exposed on the service worker global. The rationale is the lifetime of a servic worker global hinges on the UAs decision and designed to be short unless needed. It can be exposed in the future when devs request with compelling use cases.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1928,6 +1928,7 @@ This section describes a supplemental interface to the URL specification [[URL]]
 and presents methods for <a>Blob URL</a> creation and revocation.
 
 <pre class="idl">
+[Exposed=Window,DedicatedWorker,SharedWorker]
 partial interface URL {
   static DOMString createObjectURL(Blob blob);
   static DOMString createFor(Blob blob);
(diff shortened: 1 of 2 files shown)


petkaantonov:
> The rationale is the lifetime of a servic worker global hinges on the UAs decision and designed to be short unless needed. It can be exposed in the future when devs request with compelling use cases.

I really need persistent blob urls and so do others but I am ignored. See https://github.com/w3c/FileAPI/issues/27

arunranga:
Can you resubmit this request after resolving conflicts?

jungkees:
@mkruisselbrink I made this PR rebased on your latest commit. Please take a look.

/cc @arunranga

PR #34, opened 2016-03-20
Update Typed Array spec URL

goelakash:


annevk:
The fix here would be to not reference that specification at all anymore, since it's obsolete. File API should probably entirely rely on IDL for this.

arunranga:
What is the normative definition for ArrayBuffer?

annevk:
@arunranga for the purposes of most platform specifications it's IDL (which in turn will defer to JavaScript).

domenic:
This is not fixed and should be reopened.

goelakash:
@domenic I am not sure what to do here, that's why I closed it.

domenic:
Ah yes, a new issue would be good.

PR #42, opened 2016-07-11
Improve readability of Blob-related statements

izgzhen:
There are several small fixes, and I'd like to discuss about these changes (and improve it if necessary) before some of them are applied.

**line 400**: I suppose that it is not clear about what does "ignore" here mean, so I emphasize its motivation.

**line 500 and related**: I moved one note after the `Attributes` section of `Blob` and another note duplicated in two places just below the `type` attribute definition since the two notes are both `type`-related.

For the duplicated note, I deleted its appearance in two other algorithm descriptions because this note addresses what does "parsable MIME type" mean, while the two algorithms don't refer to that particular definition. So I think it is more proper to put it under `type` attribute definition (in which "parsable MIME type" is referred).

Also, I rewrote this note in a more readable form as I suppose.

**line 520**: `upto` is a typo as I suppose.

**line 610**: I rewrote description of `close()` a more readable form as I suppose.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -390,7 +390,8 @@ user agents must run the following steps:
     a copy of the bytes held by the buffer source</a>, and append those bytes to <var>bytes</var>.
   4. If <var>element</var> is a {{Blob}},
     append the bytes it represents to <var>bytes</var>.
-    The {{Blob/type}} of the {{Blob}} array element is ignored.
+    The {{Blob/type}} of the {{Blob}} array element is ignored and will not effect {{Blob/type}} of returned
+    {{Blob}} object.
 
 5. If the {{BlobPropertyBag/type}} member of the optional <code>options</code> argument is provided
   and is not the empty string,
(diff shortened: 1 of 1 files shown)


izgzhen:
Not very clear about what to do with `.html` and the failed check (Is this related to https://github.com/w3c/FileAPI/blob/gh-pages/CONTRIBUTING.md ?)

index.bs
```diff
@@ -611,8 +607,8 @@ It must act as follows:
 <h4 id="close-method">
 The close method</h4>
 
-The <dfn id="dfn-close" method for=Blob>close()</dfn> method is said to <a lt="closed" for=Blob>close</a> a {{Blob}},
-and must act as follows on the {{Blob}} on which the method has been called:
+The <dfn id="dfn-close" method for=Blob>close()</dfn> method is said to <a lt="closed" for=Blob>close</a>
+the {{Blob}} context object when it acts on the {{Blob}} as follows:
```
mkruisselbrink: I don't really like this rephrasing, in particular using "when" to combine the two sentence fragments sounds wrong to me (but I'm not a native english speaker). I'd rather just get rid of the whole "on the {{Blob}} on which the method has been called" part, but otherwise keep the current text.

mkruisselbrink:
> Not very clear about what to do with .html and the failed check (Is this related to https://github.com/w3c/FileAPI/blob/gh-pages/CONTRIBUTING.md ?)

It's fine to leave updating the .html to me. Not sure what's up with the CLA check either. This change is probably non-substantive enough for that not too matter too much, but I'm not sure what the check exactly is trying to say and what to do about it either.

index.bs
izgzhen: +1 for dropping it.

izgzhen: Fixed

dontcallmedom:
Marked as non-substantive for IPR from ash-nazg.

Issue #39, opened 2016-04-21
Re-add FileReader.readAsBinaryString() ?


inexorabletash:
Stats in Chrome: https://www.chromestatus.com/metrics/feature/timeline/popularity/814

Hovering around 0.0008% which is low. But an HTTPArchive search turned up thousands of uses; spot-checking, it's not just one library but several different uses. The method is implemented in Edge, FF,  Chrome and Safari.

So... is it worth continuing to try and drop it from implementations, or should we just accept it as a wart on FileReader, document, and move on?

(As an aside, cribbing `text()` and `arrayBuffer()` from Body to Blob would be awesome...)

annevk:
If it's implemented everywhere, yes, add it back. Not worth worrying about it.

Separate issue for `text()` and `arrayBuffer()`?

gsnedders:
Note Edge implemented it after it was dropped from the spec.

PR #44, opened 2016-07-14
Re-add readAsBinaryString()

inexorabletash:
In spec drafts circa 2011 a readAsBinaryString() method was present, predating Typed Array and readAsArrayBuffer() support. Per https://github.com/w3c/FileAPI/issues/39 this is supported by all major browsers and used in a variety of libraries, so we should accept it as a web-compat necessity and specify it.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1013,6 +1013,7 @@ interface FileReader: EventTarget {
 
   // async read methods
   void readAsArrayBuffer(Blob blob);
+  void readAsBinaryString(Blob blob);
   void readAsText(Blob blob, optional DOMString label);
   void readAsDataURL(Blob blob);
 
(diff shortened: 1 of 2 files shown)


annevk:
This doesn't seem to actually define the read operation?

inexorabletash:
> This doesn't seem to actually define/update the read operation?

https://w3c.github.io/FileAPI/#readOperationSection is used for all of the readXXX() methods. Can you clarify what is missing and/or needs updating?

index.bs
```diff
@@ -1258,6 +1264,41 @@ the user agent must run the steps below.
...
+3. Otherwise set {{FileReader/readyState}} to {{FileReader/LOADING}}.
+4. Initiate an <a>annotated task read operation</a> using the <code>blob</code> argument as <var>input</var>
+  and handle <a lt="queue a task">tasks queued</a> on the <a>file reading task source</a> per below.
+5. To <a>process read error</a> with a <a>failure reason</a>,
+  proceed to the [[#dfn-error-steps]].
+6. To <a>process read</a> <a>fire a progress event</a> called {{loadstart}} at the <a>context object</a>.
+7. To <a>process read data</a> <a>fire a progress event</a> called {{progress}} at the <a>context object</a>.
+8. To <a>process read EOF</a> run these substeps:
+  1. Set {{FileReader/readyState}} to {{FileReader/DONE}}
+  2. Set the {{FileReader/result}} attribute to the <a>body</a> returned by the <a>read operation</a> as a <a>binary string</a>.
```
annevk: I found this particularly confusing. Having this be some kind of operation that ends up converting bytes to a string would make it clearer. Currently it seems to invoke a non-normative description of "binary string"...

index.bs
```diff
@@ -1543,6 +1585,32 @@ the following steps must be followed:
...
+2. If the <code>blob</code> has been closed through the {{Blob/close()}} method,
+  throw an {{InvalidStateError}} exception
+  and <a>terminate this algorithm</a>.
+3. Otherwise, initiate a <a>read operation</a> using the <code>blob</code> argument,
+  and with the <a>synchronous flag</a> <em>set</em>.
+  If the <a>read operation</a> returns failure,
+  throw the appropriate exception as defined in [[#dfn-error-codes]].
+  <a>Terminate this algorithm</a>.
+4. If no error has occurred, return the <var>result</var> of the
+  <a>read operation</a> as an <a>binary string</a>.
```
annevk: Same here. I actually thought "binary string" was an argument to "read operation" when reading this initially. A comma would help I think, though it's still not great that "binary string" is defined in some non-normative description of a method.

inexorabletash:
This CL matches the style for the other methods. I agree they should be improved, but ISTM that should be done separately so they are consistent.

https://w3c.github.io/FileAPI/#dfn-result has:

> The list below is normative for the result attribute and is the conformance criteria for this attribute

... and then proceeds to define how the various methods generate result types, which (with this CL) defines binary string. And that section could also be improved, but again this CL just matches the existing style.

annevk:
Mkay... The existing style is not great.

mkruisselbrink:
Agreed that the way these read operations are defined isn't great, and that it would be nice to have a more explicit definition of "binary string". I filed #45 to reword these things.

PR #53, opened 2016-11-15
Remove Jonas Sicking from Editor list

vsesh:


annevk:
You should add him to the Ack section then. Cannot just erase the name completely.

PR #61, opened 2017-01-30
Fix several issues with the "Serialization of a Blob URL" algorithm.

### LINKED ISSUE #7: Use ASCII origin in blob URLs
URLs are typically ASCII in final form. Creating URLs that are not ASCII in final form is bad news and would require ugly decoding and such.

### LINKED ISSUE #58: Use HTML's definition of "Unicode serialization of an origin"
https://html.spec.whatwg.org/multipage/browsers.html#unicode-serialisation-of-an-origin

### LINKED ISSUE #59: "Unicode Serialization of a Blob URL" algorithm makes no sense
It is intended to create a new Blob URL, and then says

> Let *O* be the Blob URL’s [origin](https://url.spec.whatwg.org/#concept-url-origin).

before there even is a Blob URL.

mkruisselbrink:
This fixes #7: Use ASCII origin rather than unicode origin, matching
implementations:

This also fixes #59 by correctly getting the origin from the settings
object (but more cleanup in the whole blob URL area is needed.

And this also fixes #58.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1647,7 +1647,7 @@ a <a for=url>path</a> with one entry comprised of a UUID [[RFC4122]]
 (see [[#ABNFUUID]]),
 and an <a for=url>object</a> of the associated {{Blob}} or {{File}}.
 A Blob URL may contain an optional <a for=url>fragment</a>.
-The Blob URL is serialized as a string according to the <a>Unicode Serialization of a Blob URL</a> algorithm.
+The Blob URL is serialized as a string according to the <a>Serialization of a Blob URL</a> algorithm.
 
 A <a for=url>fragment</a>, if used,
 has a distinct interpretation depending on the media type of the {{Blob}} or {{File}} resource in question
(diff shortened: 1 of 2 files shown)


mkruisselbrink:
@annevk, @Ms2ger does this seem reasonable as a first step of fixing some of these issues?

index.bs
```diff
@@ -1709,14 +1706,15 @@ and/or <code>URL.{{URL/createFor()}}</code>:
...
-6. Return |result|.
+1. Append the ":" (U+003A COLON) character to |result|.
+1. Let |settings| be the [=incumbent settings object=]
+1. Let |origin| be |settings|'s [=environment settings object/origin=].
+1. Let |serialized| be the <a lt="ASCII serialization of an origin">ASCII serialization</a> of |origin|.
+1. If |serialized| is "null", set it to an implementation-defined value.
+1. Append |serialized| to |result|.
+1. Append the "/" character (U+0024 SOLIDUS) to |result|.
+1. Generate a UUID [[RFC4122]] as a Unicode string and append it to |result|.
+1. Return |result|.
```
annevk: Do you have tests for this algorithm somewhere? I'm curious what implementations actually do.

Ms2ger:
I don't have an opinion re: ASCII vs unicode serialization, but this does look like it improves the clarity of the prose.

index.bs
mkruisselbrink: https://github.com/w3c/web-platform-tests/pull/4657 has some tests for at least the origin serialization part (and at least firefox and chrome both ascii serialize the origin, as expected). Haven't actually added tests to check the rest of the format of these urls.

domenic:
I suspect implementations do not actually use incumbent here, so I'll sign up to write tests and submit a follow up PR if the tests bear out my suspicions. For now leaving it as incumbent seems fine.

mkruisselbrink:
> I suspect implementations do not actually use incumbent here, so I'll sign up to write tests and submit a follow up PR if the tests bear out my suspicions. For now leaving it as incumbent seems fine.

Awesome. Yeah, I have no idea if browsers are currently using incumbent or not, but it's what the spec was already saying (https://w3c.github.io/FileAPI/#originOfBlobURL), so I just kept that for now.

PR #62, opened 2017-01-31
Remove the URL.createFor method.

### LINKED ISSUE #57: Remove createFor()
Per https://github.com/w3c/mediacapture-main/issues/404 it's not implemented everywhere and the blob URL concept is not great anyway so we should get rid of this. Let's improve what we have, but not add more baggage.

mkruisselbrink:
This fixes #57

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1203,14 +1203,13 @@ the user agent must run the steps below:
 Blob Parameters</h5>
 
 The <a>asynchronous read methods</a>,
-the <a>synchronous read methods</a>,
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1203,14 +1203,13 @@ the user agent must run the steps below:
 Blob Parameters</h5>
 
 The <a>asynchronous read methods</a>,
-the <a>synchronous read methods</a>,
+the <a>synchronous read methods</a> and
```
annevk: Add an Oxford comma?

Issue #63, opened 2017-02-03
"origin of a blob URL" is redundant/confusing


domenic:
> The origin of a Blob URL
must be the same as the origin specified by the incumbent settings object at the time the method that created it was called.

This is stated as a normative statement, implying to me that blob URL's have some special origin that must be set at creation time.

However, the origin of a blob URL is instead normatively determined via [the URL Standard](https://url.spec.whatwg.org/#concept-url-origin). So this is just a statement of fact about what happens to be true according to the below algorithm, which is what creates blob URLs.

So it should be changed to a statement of fact, instead of a "must":

> The origin of a Blob URL **will** be the same as the origin specified by the incumbent settings object at the time the method that created it was called.

Furthermore, this would be much clearer if instead of defining the new term "origin of a Blob URL", it just linked the first "origin" in this sentence to https://url.spec.whatwg.org/#concept-url-origin

mkruisselbrink:
It's definitely confusing. Unfortunately currently the way the URL standard defines the origin of a blob URL does not always return the correct value, in particular for opaque origins (see whatwg/url#127). I think what we'll end up doing is store the origin of a blob URL in the Blob URL store and then using that from the URL standard to return the correct origin. But yes, there is lots of confusing, redundant and conflicting stuff with regard to blob URLs.

PR #67, opened 2017-02-09
Stop using incumbent settings object

### LINKED ISSUE #60: Stop using the "incumbent settings object"
See https://github.com/whatwg/html/issues/1430 for background.

domenic:
Fixes #60. Also adds a warning about #63.

I was unable to generate the index.html due to some macro REPOSITORYURL not being present. As such I'm not sure the warning stuff turned out all right. Maybe it would be a good idea to add instructions or a Makefile.

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1677,8 +1677,13 @@ Origin of Blob URLs</h4>
 <a>Blob URLs</a> are created using <code>URL.{{URL/createObjectURL()}}</code>,
 and are revoked using <code>URL.{{URL/revokeObjectURL()}}</code>.
 The <dfn lt="origin of a Blob URL|Blob URL's origin">origin of a Blob URL</dfn> must be the same as the [=environment settings object/origin=]
-specified by the <a>incumbent settings object</a>
+specified by the <a>current settings object</a>
 at the time the method that created it was called.
+
+WARNING: there is currently some confusion between the generic definition of the
+[=url/origin|origin of a URL=] and the specific definition of the [=origin of a Blob URL=]. This is
+tracked in <a href="https://github.com/w3c/FileAPI/issues/63">issue #63</a> and in
+<a href="https://github.com/whatwg/url/issues/127">whatwg/url#127</a>.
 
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -1677,9 +1677,14 @@ Origin of Blob URLs</h4>
 <a>Blob URLs</a> are created using <code>URL.{{URL/createObjectURL()}}</code>,
 and are revoked using <code>URL.{{URL/revokeObjectURL()}}</code>.
 The <dfn lt="origin of a Blob URL|Blob URL's origin">origin of a Blob URL</dfn> must be the same as the [=environment settings object/origin=]
-specified by the <a>incumbent settings object</a>
+specified by the <a>current settings object</a>
 at the time the method that created it was called.
 
+WARNING: there is currently some confusion between the generic definition of the
```
mkruisselbrink: I don't think "WARNING" is a thing with special meaning in bikeshed. Probably `Advisement:` or `Issue:` makes the most sense?

mkruisselbrink:
> I was unable to generate the index.html due to some macro REPOSITORYURL not being present. As such I'm not sure the warning stuff turned out all right. Maybe it would be a good idea to add instructions or a Makefile.

I'm curious what error you're getting? As far as I know I'm not doing anything special/not special configuration should be needed...

domenic:
> I'm curious what error you're getting? As far as I know I'm not doing anything special/not special configuration should be needed...

When I do

```
curl https://api.csswg.org/bikeshed/ -F file=@index.bs > index.html
```

I get an index.html that contains

```
Error running preprocessor, returned code: 1. \033[1;31mFATAL ERROR:\033[0m Found unmatched text macro [REPOSITORYURL]. Correct the macro, or escape it with a leading backslash. \033[7;31m âœ˜ \033[0m Did not generate, due to fatal errors
```

mkruisselbrink:
Ah, it's probably the fact that there is no "Repository:" metadata listed, instead relying on it being run from within a github checkout. Not sure why bikeshed even lets you do that...

mkruisselbrink:
Thanks!

Issue #49, opened 2016-08-28
Update feedback info


LJWatson:
The spec points to public-webapps@w3.org as the place to provide feedback.  This should now point to Github issues instead.

vsesh:
#69

mkruisselbrink:
@LJWatson See also my comment in #69, but I'm not sure what exactly you want to see changed? Anyway, the spec currently just has whatever the bikeshed webplatform working group templates provide for feedback text, so whatever needs changing probably needs changing in those templates?

PR #69, opened 2017-02-16
Update feedback info

vsesh:


TR.html
```diff
@@ -141,7 +141,7 @@ <h2 id="status">Status of this Document</h2>
             This document is intended to become a <abbr title="World Wide Web Consortium">W3C</abbr> Recommendation.
           
             If you wish to make comments regarding this document, please send them to 
-            <a href="mailto:public-webapps@w3.org">public-webapps@w3.org</a> 
+             <a href="https://github.com/w3c/FileAPI/issues" target="_blank">https://github.com/w3c/FileAPI/issues</a>
```
annevk: No target attributes please. No standard is doing that.

annevk: Also, ideally we put this information clearly in the heading. See e.g. https://url.spec.whatwg.org/.

vsesh: "File a bug/issue" link already in header

mkruisselbrink:
TR.html is not the file you want to be changing (possibly that file shouldn't even exist in the github repo? I have no idea how the actual publishing process works though).

But any changes should be made to the editors draft in the index.bs file (and then possibly run bikeshed to regenerate index.html), to update the index.html file. But even then most of the feedback info is taken from the webplatform working group templates bikeshed provides (such as https://github.com/tabatkins/bikeshed/blob/master/bikeshed/boilerplate/webplatform/status-ED.include).

So I'm not entirely sure what changes #49 is expecting to make. Possibly changing the "Feedback:" line at the top of the spec to not mention the mailing list. But that would possibly even require code changes to bikeshed? Not sure. It seems like bikeshed just always includes the mailing list there if there is one defined for the spec (and the default templates for the webplatform working group define the mailing list).

So sorry about this, but I'm pretty sure that whatever changes #49 wants to have made should really be made in the upstream webplatform working group bikeshed templates and/or code.

PR #77, opened 2017-04-28
Export "Blob URL Store" definition

TimothyGu:
This term is used in the [URL Standard](https://url.spec.whatwg.org/#url-parsing).

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1883,7 +1883,7 @@ developers should pair it with a corresponding call to <code>URL.{{revokeObjectU
 Lifetime of Blob URLs</h3>
 
 A global object which exposes <code>URL.{{URL/createObjectURL()}}</code>
-must maintain a <dfn id="BlobURLStore">Blob URL Store</dfn>
+must maintain a <dfn id="BlobURLStore" export>Blob URL Store</dfn>
 which is a list of <a>Blob URLs</a> created by the <code>URL.{{createObjectURL()}}</code> method,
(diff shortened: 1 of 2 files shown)


TimothyGu:
Marked as non-substantive for IPR from ash-nazg.

TimothyGu:
I marked this as non-substantive, but apologies if this process is usually done by a WG member rather than the PR author like myself.

annevk:
Wow, seems like a major security hole in the tooling. Can you also generate the HTML output? Then I can commit it.

TimothyGu:
@annevk Done.

annevk:
Thanks for doing this by the way. This should be indexed within 24h.

tripu:
@annevk, anyone can *refresh validation* and *mark as non-substantive* any PR that is not accepted yet. eg, the &ldquo;ipr details&rdquo; that are visible on WICG/animation-worklet#44 and WICG/budget-api#12 take you to the corresponding repo-manager page, where any user user can perform those actions. Chairs can still decide whether that makes sense or not, request further IPR commitments, etc.

annevk:
That seems like a very strange model.

tripu:
@annevk: I imagine @darobin adopted the simple/optimistic approach to security, and erred on the side of letting users do sensible things by default, instead of restricting actions too much.

annevk:
There's no security with this model though. You might as well not have the check, since everything depends on the chairs.

PR #81, opened 2017-05-11
Replace RFC1738 with RFC4266 (gopher)

foolip:
Bikeshed said "Obsolete biblio ref: [rfc1738] is replaced by [rfc4266]."


foolip:
I really just wanted to regenerate index.html, but couldn't because of the error :)

PR #90, opened 2017-11-29
Add BlobPropertyBag.endings back to the spec.

### LINKED ISSUE #46: Document BlobPropertyBag "endings", or keep pushing to remove?
https://jsfiddle.net/422r7n3z/ - exercises `new Blob([s], {endings:e}).size`

On Windows, in Chrome/Firefox:
transparent + "abc\r\n" = 5
transparent + "abc\n" = 4
transparent + "abc\r" = 4
native + "abc\r\n" = 5
native + "abc\n" = 5
native + "abc\r" = 5

On Windows, in IE11:
transparent + "abc\r\n" = 5
transparent + "abc\n" = 4
transparent + "abc\r" = 4
native + "abc\r\n" = 5
native + "abc\n" = 4
native + "abc\r" = 4

On OSX and Linux, in Chrome/Firefox/Safari:
transparent + "abc\r\n" = 5
transparent + "abc\n" = 4
transparent + "abc\r" = 4
native + "abc\r\n" = 4
native + "abc\n" = 4
native + "abc\r" = 4

(I have not tested Opera or Edge; I assume those match Chrome and IE11)

Note that all 4 browsers - _including IE which does not appear to respect the option otherwise_ - throw if an invalid option is passed as `endings`. 

I gathered usage stats in Chrome stable: https://www.chromestatus.com/metrics/feature/timeline/popularity/1320 - it's 0% to several significant figures (people running web-platform-tests would account for the absolute #s).

Should we proceed with removing, or does anyone want to advocate for keeping it? @smaug----, @foolip, @aliams, .... ?

mkruisselbrink:
This was removed 5 years ago, but browsers still implement it so bring
it back to the spec.

This fixes #46.
Tests in w3c/web-platform-tests#8456

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -49,6 +49,7 @@ spec: html
 spec: infra
   type: dfn
     text: list
+    text: string
 spec: streams
   type: dfn
     text: chunk
@@ -230,8 +231,11 @@ interface Blob {
             optional DOMString contentType);
 };
 
+enum EndingTypes {"transparent", "native"};
+
 dictionary BlobPropertyBag {
   DOMString type = "";
+  EndingTypes endings = "transparent";
 };
 
 typedef (BufferSource or Blob or USVString) BlobPart;
@@ -253,6 +257,7 @@ Their [=deserialization step=], given |serialized| and |value|, are:
 <h3 id="constructorBlob">
 Constructors</h3>
 
+<div algorithm="blob-constructor">
 The {{Blob()}} constructor can be invoked with zero or more parameters.
 When the {{Blob()}} constructor is invoked,
 user agents must run the following steps:
@@ -262,41 +267,22 @@ user agents must run the following steps:
   with {{Blob/size}} set to 0,
   and with {{Blob/type}} set to the empty string.
 
-2. Otherwise, the constructor is invoked with a {{blobParts}} sequence.  Let |a| be that sequence.
+1. Let |bytes| be the result of [=processing blob parts=] given {{blobParts}} and {{Blob/Blob(blobParts, options)/options}}.
 
-3. Let |bytes| be an empty sequence of bytes.
-
-4. Let |length| be |a|'s length.
-  For 0 ≤ |i| &lt; |length|, repeat the following steps:
-
-  1. Let |element| be the |i|th element of |a|.
-  2. If |element| is a {{USVString}}, run the following substeps:
-
-    1. Append the result of [=UTF-8 encoding=] |s| to |bytes|.
-
-      Note: The algorithm from WebIDL [[WebIDL]] replaces unmatched surrogates in an invalid utf-16 string
-      with U+FFFD replacement characters.
-      Scenarios exist when the {{Blob}} constructor may result in some data loss
-      due to lost or scrambled character sequences.
-  3. If |element| is a {{BufferSource}}, <a lt="get a copy of the buffer source">get
-    a copy of the bytes held by the buffer source</a>, and append those bytes to |bytes|.
-  4. If |element| is a {{Blob}},
-    append the bytes it represents to |bytes|.
-    The {{Blob/type}} of the {{Blob}} array element is ignored and will not affect {{Blob/type}} of returned
-    {{Blob}} object.
-
-5. If the {{BlobPropertyBag/type}} member of the optional <code>options</code> argument is provided
+1. If the {{BlobPropertyBag/type}} member of the optional {{Blob/Blob(blobParts, options)/options}} argument is provided
   and is not the empty string,
   run the following sub-steps:
 
   1. Let |t| be the {{BlobPropertyBag/type}} dictionary member.
     If |t| contains any characters outside the range U+0020 to U+007E,
     then set |t| to the empty string and return from these substeps.
-  2. Convert every character in |t| to [=ASCII lowercase=].
+  1. Convert every character in |t| to [=ASCII lowercase=].
 
-6. Return a {{Blob}} object referring to |bytes| as its associated <a>byte</a> sequence,
+1. Return a {{Blob}} object referring to |bytes| as its associated <a>byte</a> sequence,
   with its {{Blob/size}} set to the length of |bytes|,
   and its {{Blob/type}} set to the value of |t| from the substeps above.
+
+</div>
 
 <h4 id="constructorParams">
 Constructor Parameters</h4>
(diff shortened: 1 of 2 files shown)


index.bs
```diff
@@ -308,15 +294,95 @@ The {{Blob()}} constructor can be invoked with the parameters below:
...
+</div>
+
+<div algorithm="convert-line-endings-to-native">
+To <dfn lt="convert line endings to native|converting line endings to native">
+convert line endings to native</dfn> in a [=string=] |s|,
+run the following steps:
+
+1. Let |native line ending| be be the [=code point=] U+000A LF,
+  or the [=code point=] U+000D CR followed by the [=code point=] U+000A,
+  as determined by the underlying platform's conventions.
```
annevk: Maybe also say that if there are different conventions it has to be LF?

annevk: E.g., by setting it to LF and only overwriting it to CR LF if the platform conventions are that.

index.bs
```diff
@@ -308,15 +294,95 @@ The {{Blob()}} constructor can be invoked with the parameters below:
...
+  1. If the [=code point=] at |position| within |s| equals U+000A LF,
+    advance |position| by 1 and append |native line ending| to |result|.
+
+  1. Let |token| be the result of [=collecting a sequence of code points=]
+    that are not equal to U+000A LF or U+000D CR
+    from |s| given |position|.
+
+  1. Append |token| to |result|.
+
+1. Return |result|.
```
annevk: I think it would be okay to say something like "Replace any occurrence of LF, CR LF, or CR with _native_."

It seems your algorithm doesn't consider a lone CR a line ending. I'd assume browsers do treat it as such. Is that tested?

index.bs

index.bs
mkruisselbrink: Good point, I more or less copied the algorithm from https://www.w3.org/TR/2012/WD-FileAPI-20121025/#convenienceAPI, but definitely should also convert a lone CR to native.

mkruisselbrink: Not sure I like the shorter phrasing though. I think being explicit is good in this case since both the search strings are subsets of each other, and the replacement string is also part of the search strings, so there is plenty of possibly ambiguity in what exactly it means to do such a search/replace operation.

mkruisselbrink: Of course it might make sense to define such a string replace operation in the infra spec, similar to how split operations are already defined there...

annevk: Up to you really. There's probably some things to abstract here as there's many places dealing with LF/CRLF/CR, but no rush.

index.bs
```diff
@@ -308,15 +294,103 @@ The {{Blob()}} constructor can be invoked with the parameters below:
   <dd>which takes any number of the following types of elements, and in any order:
     * {{BufferSource}} elements.
     * {{Blob}} elements.
-    * {{DOMString}} elements.
+    * {{USVString}} elements.
 
   <dt id="dfn-BlobPropertyBagMembers">An *optional* {{BlobPropertyBag}}
-  <dd>which takes one member:
+  <dd>which takes two members:
```
inexorabletash: Maybe "these optional members" ?

index.bs
```diff
@@ -308,15 +294,103 @@ The {{Blob()}} constructor can be invoked with the parameters below:
...
+
+1. Let |bytes| be an empty sequence of bytes.
+
+1. For each |element| in |parts|:
+
+  1. If |element| is a {{USVString}}, run the following substeps:
+
+    1. Let |s| be |element|.
+
+    1. If the {{BlobPropertyBag/endings}} member of |options| is set to {{"native"}},
```
inexorabletash: nit: I think "set" is a verb in spec language, so "is set to" is used during initialization, not testing. Could just be `|options| is {{"native"}}` here.

index.bs
```diff
@@ -308,15 +294,103 @@ The {{Blob()}} constructor can be invoked with the parameters below:
...
+</div>
+
+<div algorithm="convert-line-endings-to-native">
+To <dfn lt="convert line endings to native|converting line endings to native">
+convert line endings to native</dfn> in a [=string=] |s|,
+run the following steps:
+
+1. Let |native line ending| be be the [=code point=] U+000A LF.
+
+1. If it matches the underlying platform's conventions,
```
inexorabletash: "it" reads ambiguously here in English (in a naive reading it refers to the previous step's context)

Better would be something phrased as: "If the underlying platform's conventions are ... then ... "

(And the clause might be "to represent newlines as a carriage return and line feed sequence" but I'm more concerned about the sentence structure)

index.bs
```diff
@@ -308,15 +294,103 @@ The {{Blob()}} constructor can be invoked with the parameters below:
...
+  set |native line ending| to the [=code point=] U+000D CR
+  followed by the [=code point=] U+000A.
```
inexorabletash: nit: Previous line has `U+000D CR` but this line has only `U+000A`

https://infra.spec.whatwg.org/#code-points says to use the name when it's not printable.

Super nitpicky: per http://www.unicode.org/charts/PDF/U0000.pdf it would be **U+000D CARRIAGE RETURN (CR)** and **U+000A LINE FEED (LF)** although Infra itself just uses LF

index.bs
```diff
@@ -308,15 +294,103 @@ The {{Blob()}} constructor can be invoked with the parameters below:
...
+
+1. Let |token| be the result of [=collecting a sequence of code points=]
+  that are not equal to U+000A LF or U+000D CR
+  from |s| given |position|.
+
+1. Append |token| to |result|.
+
+1. While |position| is not past the end of |s|:
+
+  1. If the [=code point=] at |position| within |s| equals U+000D CR:
```
inexorabletash: Subsequent steps would be simpler if you assign a local variable (e.g. _cp_) to _code point at position within s_

index.bs
```diff
@@ -308,15 +294,103 @@ The {{Blob()}} constructor can be invoked with the parameters below:
...
+
+    1. Append |native line ending| to |result|.
+
+    1. Advance |position| by 1.
+
+    1. If |position| is not past the end of |s|
+      and the [=code point=] at |position| within |s| equals U+000A LF
+      advance |position| by 1.
+
+  1. If the [=code point=] at |position| within |s| equals U+000A LF,
```
inexorabletash: Needs an "Otherwise" (or to handle position being past the end)

index.bs
annevk: I would prefer the shorter variant (we use those in Infra).

index.bs
```diff
@@ -230,8 +231,11 @@ interface Blob {
             optional DOMString contentType);
 };
 
+enum EndingTypes {"transparent", "native"};
```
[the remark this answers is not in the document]

index.bs
mkruisselbrink: done

index.bs

index.bs
mkruisselbrink: Yeah, I didn't like that sentence structure either, but wasn't quite sure what clause to use otherwise. You're suggestion works, thanks.

index.bs
mkruisselbrink: Agreed that the shorter variant is clear enough, and seems to match what infra suggests. Added here, which I believe was the only place I was missing the name.

index.bs
mkruisselbrink: Added an otherwise.

index.bs
mkruisselbrink: Not sure about that. It's only used twice (at least with the Otherwise added below, else it would only be referenced once) so I don't think we gain much by adding a temporary local variable.

index.bs
mkruisselbrink: Okay, I think I'll leave it for future cleanups to decide what might make sense to move into infra.

Issue #89, opened 2017-11-08
Clarify Lifetime of BlobStore


bmeck:
I wrote up [an example](https://jsbin.com/yomodiz/1/edit?js,console) trying to figure out how long browsers allow you to load from a given Realm's BlobStore.

It creates a new URL inside of an `<iframe>` and then removes the `<iframe>` from the DOM and tries to use that URL to load a 2nd `<iframe>`.

The behavior seems to be inconsistent:

|Browser|Version|Fires 2nd `<iframe>`'s `onload`|Fires 2nd `<iframe>`'s `onerror`|
|----|----|----|----|
|Safari|11|No|No|
|Firefox|52|No|Yes|
|Chrome|61|Yes|No|

I think the spec is a bit iffy on timeline of when URLs become unavailable for load.

mkruisselbrink:
Yeah, unfortunately the spec is a bit of a mess around the whole Blob URL Store concept (as you point out it sort of is per global, but then there are examples that show two different (same origin) windows "sharing" the same Blob URL Store).

That said, there is more subtlety going on here, and I'm not sure your example is entirely correct. First of all I'm not sure how you're detecting "post GC"? I don't think we really spec when a document is garbage collected, just what should happen when it happens. And then another issue with your code is that you seem to (synchronously) create and insert the new iframe after removing the old iframe. This insertion then should (also synchronously) resolve the blob URL (as part of parsing the src attribute), which should resolve the blob URL, and thus loading should generally succeed, even if removing the iframe actually would revoke the URL (since that involves GC and thus almost certainly isn't done synchronously).

So yes, lots of room for improvement with how the Blob URL Store is specified, and this is definitely something I hope to work on this quarter (since I'm also working on refactoring parts of the chrome implementation, and running into all kinds of edge cases that aren't clearly specified in the current spec).

annevk:
So given that browsers allow users to open blob URLs in a new top-level tab, it seems to me that the store is effectively per browser and you can't have any narrower scope. We can place some restrictions here and there on usage, but as long as top-level tabs are game they all seem artificial to me (other than security restrictions of course).

mkruisselbrink:
[...]

Issue #20, opened 2015-10-19
Fix Blob Origin


arunranga:
This issue is to fix Blob URL's origin, based largely on bugs filed on HTML.

Currently, File API has this:

"The origin of a Blob URL must be the same as the effective script origin specified by the incumbent settings object at the time the method that created it -- either URL.createObjectURL or URL.createFor -- was called."

This should be much better.

PR #92, opened 2017-12-04
Refactor Blob URL Store definition.

mkruisselbrink:
Still far from done, but a start at trying to better specify how Blob URLs work. Still need to write tests, update URL and Fetch specs, clarify a bunch of stuff in here, and make sure the described behavior is actually correct...

This should ultimately address #89, #63, and #20.


index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
 
 This section defines a scheme for a URL used to refer to {{Blob}} objects (and {{File}} objects).
```
domenic: Maybe do `[=URL/scheme=]` and `[=URL=]` here.

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
+[=Blob URLs=] can also be used to navigate to as well as to trigger downloads of
+locally generated data.
+
+For this purpose two static methods are exposed on the {{URL}} interface,
+{{URL/createObjectURL(blob)}} and {{URL/revokeObjectURL(url)}}.
+The first method creates a mapping from a URL to a Blob,
+and the second method revokes said mapping.
+As long as a mapping exist a Blob  can't be garbage collected,
+so some care must be taken to revoke a URL as soon as a reference is no longer needed.
+All URLs are revoked when the global that created the URL itself goes away.
```
domenic: In this paragraph, `Blob` should be `{{Blob}}`, I think.

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
+This enables integration of {{Blob}}s and {{File}}s with other Web Platform APIs
+that are only designed to be used with URLs, such as the <{img}> element.
+[=Blob URLs=] can also be used to navigate to as well as to trigger downloads of
+locally generated data.
+
+For this purpose two static methods are exposed on the {{URL}} interface,
+{{URL/createObjectURL(blob)}} and {{URL/revokeObjectURL(url)}}.
+The first method creates a mapping from a URL to a Blob,
+and the second method revokes said mapping.
+As long as a mapping exist a Blob  can't be garbage collected,
```
domenic: "As long as the mapping exists the Blob can't...", "...revoke the URL...", "as soon as the reference..."

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
+so some care must be taken to revoke a URL as soon as a reference is no longer needed.
+All URLs are revoked when the global that created the URL itself goes away.
+
+<h3 id="url-model">
+Model</h3>
+
+Each user agent must maintain a <dfn id="BlobURLStore" export>Blob URL Store</dfn>.
```
domenic: Nit: I think generally we don't capitalize terms defined this way, so e.g. "blob URL store" and "blob URL entry" would be more conventional.

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
+A <dfn>Blob URL Entry</dfn> consists of
+an <dfn for="Blob URL Entry">object</dfn> (typically a {{Blob}},
+but other specs can extend this to refer to other types of objects),
+and an <dfn for="Blob URL Entry">environment</dfn> (an [=environment settings object=]).
+
+[=map/Keys=] in the [=Blob URL Store=] (also known as <dfn lt="Blob URL" export>Blob URLs</dfn>)
+are [=valid URL strings=] that when [=URL parser|parsed=]
+result in a [=/URL=] with a [=url/scheme=] equal to <code>"blob"</code>,
+a [=url/host=] equal to the [=environment settings object/origin=]
+of the [=environment settings object=] in which the URL was create,
```
domenic: s/was create/was created

Is this the origin of the corresponding blob URL entry's environment? If so maybe I would make this correspondence explicit somewhere.

If it does *not* correspond, then an example of when that's the case would also be useful.

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
 </div>
 
+<div algorithm="addBlobURL">
+To <dfn id="add-an-entry" lt="add an entry|add the entry|add an entry to the Blob URL Store|add the entry to the Blob URL Store|adding an entry to the Blob URL Store" for="blob url store">
+add an entry to the Blob URL Store</dfn> for a given |object|,
+run the following steps:
 
-<h4 id="originOfBlobURL">
-Origin of Blob URLs</h4>
+1. Let |store| be the user agent's [=Blob URL Store=];
```
domenic: Semicolon should be period

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
-<a href="https://github.com/whatwg/url/issues/127">whatwg/url#127</a>.
+<div algorithm="removeBlobURL">
+To <dfn id="removeTheEntry" lt="remove an entry|remove the entry|remove an entry from the Blob URL Store|remove the entry from the Blob URL Store" for="blob url store">
+remove an entry from the Blob URL Store</dfn> for a given |url|,
+run the following steps:
 
-[=CORS protocol|Cross-origin requests=] on Blob URLs must return a <a>network error</a>.
+1. Let |store| be the user agent's [=Blob URL Store=];
+1. Let |url string| be the result of [=URL serializer|serializing=] |url|.
+1. Remove |store|[|url string|].
```
domenic: Since you seem to like linking to Infra things, you can link to remove too :)

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
-which is invoked by <code>URL.{{URL/createObjectURL()}}</code>:
+<div algorithm="createBlobURL">
+To <dfn id="unicodeBlobURL" lt="generate a new Blob URL|generating a new Blob URL">
+generate a new Blob URL</dfn>, run the following steps:
 
 1. Let |result| be the empty string.
-  Append the string "blob"
-  (that is, the Unicode code point sequence U+0062, U+006C, U+006F, U+0062)
+1. Append the string "blob"
+  (that is, the Unicode [=code point=] sequence U+0062, U+006C, U+006F, U+0062)
```
domenic: Strings should be "`blob`", i.e. `<code>` inside the quotes. Here and for "null" below.

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
-Note: The [[!URL]] and [[!Fetch]] specifications should be considered normative for parsing and fetching <a>Blob URLs</a>.
+<div algorithm="resolveURL">
+To <dfn for="Blob URL" lt="resolve">resolve a Blob URL</dfn> given a |url| (a [=URL=]), run the following steps:
 
-<a>Blob URLs</a> are <dfn lt="dereference">dereferenced</dfn>
-when the user agent retrieves the resource identified by the Blob URL
-and returns it to the requesting entity.
-This section provides guidance on [=/requests=] and [=/responses=]
+1. [=Assert=]: |url|'s [=url/scheme=] is <code>"blob"</code>.
+1. Let |store| be the user agent's [=Blob URL Store=];
```
domenic: semicolon should be period

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
+1. Let |url string| be the result of [=URL serializer|serializing=] |url| with the *exclude fragment flag* set.
+1. Let |entry| be the result of [=map/getting=] |store|[|url string|].
```
domenic: You can just say "Let |entry| be |store|[|url string|]." Or even "Return |store|[|url string|]."

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
-This section provides sample exchanges between web applications and user agents using <a>Blob URLs</a>.
-A [=/request=] can be triggered using HTML markup of the sort
-<code class="lang-markup">&lt;img src="blob:http://example.org:8080/550e8400-e29b-41d4-a716-446655440000"></code>.
-These examples merely illustrate the [=/request=] and [=/response=];
-web developers are not likely to interact with all the headers,
-but the {{XMLHttpRequest/getAllResponseHeaders()}} method of {{XMLHttpRequest}}, if used,
-will show relevant [=/response=] headers.
+1. [=Assert=]: |url|'s [=url/scheme=] is <code>"blob"</code>.
+1. Let |entry| be the result of [=Blob URL/resolve|resolving=] |url|.
+1. If |entry| is not undefined, return |entry|'s [=Blob URL Entry/environment=]'s [=environment settings object/origin=].
```
domenic: Hmm, so unlike JS, we don't say that getting something from a map returns undefined when it doesn't exist. Instead you should probably do something like "If map[x] exists, return map[x]; otherwise, return failure." Then test for failure here. Or null instead of failure, perhaps.

index.bs
```diff
@@ -1898,58 +1775,54 @@ unless the user agent also implements the URL [[URL]] specification.
 In other words, <code>URL.prototype</code> must evaluate to true if the user agent implements the URL [[URL]] specification,
```
domenic: This "ECMAScript user agents" section seems unnecessary these days, where everyone has a `URL` interface.

index.bs
```diff
@@ -1898,58 +1775,54 @@ unless the user agent also implements the URL [[URL]] specification.
...
+1. [=Remove an entry from the Blob URL Store=] for |url|.
+1. Let |settings| be the [=current settings object=]
+1. Let |origin| be the origin of |url|. FIXME
+1. If |origin| is not [=same origin=] with |settings|'s [=environment settings object/origin=], return.
+1. [=Remove an entry from the Blob URL Store=] for |url|.
+
+Issue: Do we actually need a same-origin check? It doesn't seem like this is something browser currently do, but it seems nice to have.
+
+Note: This means that rather than throwing some kind of error, attempting to revoke a URL that isn't registered will silently fail.
+User agents may display a message on the error console is this happens.
```
domenic: s/may/might, or make this normative

index.bs
```diff
@@ -1898,58 +1775,54 @@ unless the user agent also implements the URL [[URL]] specification.
...
+
+Note: Attempts to dereference |url| after it has been revoked will result in a [=network error=].
```
domenic: s/dereference/fetch?

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
+1. Append U+003A COLON (:) to |result|.
 1. Let |settings| be the [=current settings object=]
 1. Let |origin| be |settings|'s [=environment settings object/origin=].
 1. Let |serialized| be the <a lt="ASCII serialization of an origin">ASCII serialization</a> of |origin|.
 1. If |serialized| is "null", set it to an implementation-defined value.
 1. Append |serialized| to |result|.
-1. Append the "/" character (U+0024 SOLIDUS) to |result|.
-1. Generate a UUID [[RFC4122]] as a Unicode string and append it to |result|.
+1. Append U+0024 SOLIDUS (/) to |result|.
+1. Generate a UUID [[!RFC4122]] as a Unicode string and append it to |result|.
```
inexorabletash: nit: "Unicode string" can just be "string"

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
-that the "File Save As" user interface in user agents takes into account the {{File}}'s {{File/name}} attribute
-if presenting a default name to save the file as.
-
-<h4 id="NetworkError">
-Network Errors</h4>
-
-Responses that do not succeed with a <a href="#TwoHundredOK">200 OK</a>
-act as if a <a>network error</a> has occurred.
-Network errors are used when:
+Issue: This more or less matches chrome's behavior,
```
inexorabletash: nit: capitalize "Chrome" if this is going to land.

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
-A [=/request=] can be triggered using HTML markup of the sort
-<code class="lang-markup">&lt;img src="blob:http://example.org:8080/550e8400-e29b-41d4-a716-446655440000"></code>.
-These examples merely illustrate the [=/request=] and [=/response=];
-web developers are not likely to interact with all the headers,
-but the {{XMLHttpRequest/getAllResponseHeaders()}} method of {{XMLHttpRequest}}, if used,
-will show relevant [=/response=] headers.
+1. [=Assert=]: |url|'s [=url/scheme=] is <code>"blob"</code>.
+1. Let |entry| be the result of [=Blob URL/resolve|resolving=] |url|.
+1. If |entry| is not undefined, return |entry|'s [=Blob URL Entry/environment=]'s [=environment settings object/origin=].
+1. Let |nested url| be the result of [=URL parser|parsing=] |url|'s [=url/path=][0].
```
inexorabletash: What's the `[0]` ?

Also, isn't url/path the UUID and url/host is the serialized origin?

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
-  then a user agent acts as if a <a>network error</a> has occurred.
-</div>
+Issue: This needs a similar hook when a worker is unloaded.
 
 <h3 id="creating-revoking">
 Creating and Revoking a Blob URL</h3>
 
 <a>Blob URLs</a> are created and revoked using methods exposed on the {{URL}} object,
-supported by global objects {{Window}} [[HTML]] and {{WorkerGlobalScope}} [[Web Workers]].
+supported by global objects {{Window}} and {{WorkerGlobalScope}}.
```
inexorabletash: Is this "supported by...." sentence fragment even necessary? 

And maybe "static methods" as used above?

index.bs
```diff
@@ -1898,58 +1775,54 @@ unless the user agent also implements the URL [[URL]] specification.
...
-
-      Since <code>window1</code> and <code>window2</code> are in the <a>same origin</a>
-      and share the same <a>Blob URL Store</a>,
-      the <code>URL.{{revokeObjectURL()}}</code> call
-      ensures that subsequent dereferencing of <code>myurl</code>
-      results in a the user agent acting as if a <a>network error</a> has occurred.
-    </div>
-</dl>
+1. Let |url| be the result of [=adding an entry to the Blob URL Store=] for |blob|.
+4. Return |url|.
```
inexorabletash: nit: `1.`

index.bs
```diff
@@ -1898,58 +1775,54 @@ unless the user agent also implements the URL [[URL]] specification.
...
+
+</div>
+
+<div algorithm="revokeObjectURL">
+The <dfn method for=URL id="dfn-revokeObjectURL">revokeObjectURL(|url|)</dfn> static method must run these steps:
+
+1. [=Remove an entry from the Blob URL Store=] for |url|.
```
inexorabletash: I'm assuming these steps are still being worked on? Looks like this is incomplete. (Redundant first/last steps, FIXME, etc)

index.bs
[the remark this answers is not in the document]
annevk: You could even make this append the string "blob:" since strings are a well-defined concept now. I wonder though if this is the way we should generate them or if we should create a URL record and then serialize that.

index.bs
```diff
@@ -1598,286 +1600,161 @@ A URL for Blob and File reference</h2>
...
 
+<div algorithm="addBlobURL">
+To <dfn id="add-an-entry" lt="add an entry|add the entry|add an entry to the Blob URL Store|add the entry to the Blob URL Store|adding an entry to the Blob URL Store" for="blob url store">
+add an entry to the Blob URL Store</dfn> for a given |object|,
+run the following steps:
 
-<h4 id="originOfBlobURL">
-Origin of Blob URLs</h4>
+1. Let |store| be the user agent's [=Blob URL Store=];
+1. Let |url| be the result of [=generating a new Blob URL=].
```
annevk: You need to pass |object| as argument here presumably?

index.bs

index.bs
domenic: Hmm, I guess I misremembered Infra. Well, yeah, this definitely needs clarity; "nothing" is pretty ambiguous. I will open a new issue.

mkruisselbrink:
I started working on tests in https://github.com/w3c/web-platform-tests/pull/8612 fwiw

index.bs
mkruisselbrink: Actually after testing I decided to do keep the same-origin check, as at least chrome and firefox already seem to do that (well, Firefox seems to have a same-origin check, Chrome only lets you revoke URLs from the context that created them entirely).

mkruisselbrink:
Getting back to this (well, not really with any recent changes on this particular PR), I think I'll just go ahead and merge this (and the accompanying tests) tomorrow, working on the remaining issues (both here and in things like the URL spec) as follow-ups.

PR #103, opened 2018-06-19
Allow MediaSource, disallow any more extensions

### LINKED ISSUE #100: Include MediaSource overload for createObjectURL
Currently [the MSE spec](https://github.com/w3c/media-source/issues/211) adds overload for createObjectURL using partial interface, but [the Web IDL spec says](https://heycam.github.io/webidl/#idl-overloading):

>Operations must not be overloaded across interface, partial interface, interface mixin, and partial interface mixin definitions.

We should directly include the overload in the original definition.

saschanaz:
Closes #100.

Edit: It seems the WebPlat group only allows authorized members. Hmm.

***
<a href="https://pr-preview.s3.amazonaws.com/saschanaz/FileAPI/pull/103.html" title="Last updated on Feb 1, 2019, 9:26 AM UTC (7ab0c1e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/103/94f1192...saschanaz:7ab0c1e.html" title="Last updated on Feb 1, 2019, 9:26 AM UTC (7ab0c1e)">Diff</a>


saschanaz:
Can I just mark this as non-substantive to get a review?

marcoscaceres:
This is editorial, by the looks of it. So, yeah, go for it @saschanaz.

saschanaz:

annevk:
@tabatkins does Bikeshed's IDL parser miss some functionality? It errors on the IDL here.

saschanaz:
Woops, it's because `object` is a keyword. Replaced it by `obj`.

annevk:
Thanks @saschanaz! Sorry for the ping @tabatkins, too soon apparently.

index.bs
```diff
@@ -1730,14 +1729,14 @@ and presents methods for <a>blob URL</a> creation and revocation.
 <pre class="idl">
 [Exposed=(Window,DedicatedWorker,SharedWorker)]
 partial interface URL {
-  static DOMString createObjectURL(Blob blob);
+  static DOMString createObjectURL((Blob or MediaSource) obj);
```
Ms2ger: Is this valid? `MediaSource` is not exposed everywhere `URL` is.

mkruisselbrink: Why wouldn't it be "valid" (for whatever definition of valid you're asking about)? You just won't be able to pass in a MediaSource when in a worker, but that doesn't seem that crazy?

I'm not sure how else we'd be writing this in IDL either, since you can't have two overloads of the same method with different Exposed= values for example (perhaps you could somehow work around it with two separate partial interfaces? But if that is a possible way to get different overloads with different Exposed= values, I'm not sure why it isn't just allowed directly).

annevk: I'm pretty sure we had this discussion before (likely different people) and decided it ought to work. See also `send()` in https://xhr.spec.whatwg.org/#interface-xmlhttprequest.

dannyjr894:
Allow extentions (#103)

PR #106, opened 2018-07-23
Change label to encoding for FileReader.readAsText

### LINKED ISSUE #105: Restore name of second argument to FileReader.readAsText
In eaf8fd9d4d64a93c8b5fddcd63d48cba923ff005 ("Cleaned up wrong TOC"), arangana made a change from `readAsText(blob[, encoding])` to `readAsText(blob[, label])` but I have not been able to find a motivation.

I'm proposing that the name `encoding` is brought back in order to clarify what the parameter does, without having to read the surrounding text in the spec.

My main motivation for filing this issue is that some documentation generators depend on the wording in this spec, which trickles down an arguably poor parameter name choice to all developers using the DOM APIs. (See https://github.com/Microsoft/TSJS-lib-generator/pull/536).

The reasoning for `encoding` specifically is:

1. It's the parameter name that had already been settled on in the spec from 2009–2013
2. It's still used by respected documentation sources (eg., [MDN](https://developer.mozilla.org/en-US/docs/Web/API/FileReader/readAsText))
3. It's arguably the most concise description of what the valid input values are (eg., `"UTF-8"`)

Here is what today's documentation looks like in Visual Studio Code (TypeScript) due to the name choice in this repository:

screen shot 2018-07-21 at 12 55 53

blixt:
Along with `FileReaderSync`, of course.

Before:

    readAsText(blob[, label])

After:

    readAsText(blob[, encoding])

(I have not run the bikeshedding tool to regenerate `index.html` as I'm not sure how to.)

Fixes #105

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1137,7 +1137,7 @@ the user agent must run the steps below.
 The {{FileReader/readAsText()}} method</h5>
 
 The {{FileReader/readAsText()}} method can be called with an optional parameter,
-<dfn argument for="FileReader/readAsText(blob, encoding), FileReader/readAsText(), FileReaderSync/readAsText(blob, encoding)" id="dfn-encoding">encoding</dfn>,
+<dfn argument for="FileReader/readAsText(blob, encoding), FileReader/readAsText(), FileReaderSync/readAsText(blob, encoding)" id="dfn-encoding" oldids="dfn-label">encoding</dfn>,
 which is a {{DOMString}} argument that represents the label of an encoding [[!Encoding]];
 if provided, it must be used as part of the <a>encoding determination</a> used when processing this method call.
 
--- a/index.bs
+++ b/index.bs
@@ -953,7 +953,7 @@ interface FileReader: EventTarget {
   // async read methods
   void readAsArrayBuffer(Blob blob);
   void readAsBinaryString(Blob blob);
-  void readAsText(Blob blob, optional DOMString label);
+  void readAsText(Blob blob, optional DOMString encoding);
   void readAsDataURL(Blob blob);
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -1136,11 +1136,11 @@ the user agent must run the steps below.
 The {{FileReader/readAsText()}} method</h5>
 
 The {{FileReader/readAsText()}} method can be called with an optional parameter,
-<dfn argument for="FileReader/readAsText(blob, label), FileReader/readAsText(), FileReaderSync/readAsText(blob, label)" id="dfn-label">label</dfn>,
+<dfn argument for="FileReader/readAsText(blob, encoding), FileReader/readAsText(), FileReaderSync/readAsText(blob, encoding)" id="dfn-encoding">encoding</dfn>,
```
annevk: I'd somewhat prefer it if we kept the IDs unchanged or at least kept the old IDs working.

blixt: I can add the legacy reference too. Is there a precedent for keeping legacy links that I can follow in terms of documenting the reasoning, or possibly a Bikeshed tooling support for adding the legacy reference?

If not, I propose to add this line below:

```html
<a id="dfn-label"><!-- Legacy anchor for when argument name was "label" --></a>
```

annevk: `oldids` as attribute should work, but I'm not sure it's worth changing this at all. I guess the editor can decide.

blixt: Ah neat! I'll just add that in then since the support already exists. If for any reason (but I don't see one) it should be removed I don't mind removing it again.

mkruisselbrink:

PR #110, opened 2018-10-04
Fix "undefined" reference in parsable MIME type Note

rwlbuis:
AFAICS that algorithm can return "failure" but not "undefined".


mkruisselbrink:

mkruisselbrink:
Thanks!

PR #117, opened 2019-03-07
Add simpler reading methods to Blob interface.

### LINKED ISSUE #40: Streams are hot, FileReader.readAs____ is not
Large Blob/Files becomes an issue if you don't slice it into chunks...

Streams brings other cool feature like piping, back pressure & transform
Currently I'm using Response (part of Fetch) to transform a blob to a stream that i can "read into" step by step

```js
var stream = new Response(Blob).body
var reader = stream.getReader()
```

This works okey. but it feels hacky.
Could we get new method to get a stream from File & Blob's please?

Suggestions:

```js
blob.getReadableByteStream()
blob.getReadableByteStream().getReader()
blob.getReadableByteStream().pipeTo(anotherStream)
blob.createReadableByteStream()
```

mkruisselbrink:
This fixes #40 

For now this just more or less copies the hand-wavy text of "perform a read operation". We'll need to better specify what actually reading from a blob does. Separately in a follow-up I would also like to rephrase the existing FileReader/FileReaderSync spec text in terms of readable stream operations, being much more precise when and how state is updated and events are queued.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/pull/117.html" title="Last updated on Mar 26, 2019, 4:36 PM UTC (42f8a45)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/117/795750f...42f8a45.html" title="Last updated on Mar 26, 2019, 4:36 PM UTC (42f8a45)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -569,56 +569,46 @@ It must act as follows:
   </pre>
 </div>
 
-<h4 id="stream-method-algo">
-The stream method</h4>
+### The {{Blob/stream()}} method ### {#stream-method-algo}
 
-The <dfn method for=Blob>stream()</dfn> method returns the contents
-of the blob as a {{ReadableStream}}.
-
-It must return the result of [=getting a stream for a Blob=] for the [=context object=].
+The <dfn method for=Blob>stream()</dfn> method, when invoked, must return
+the result of [=getting a stream for a Blob=] for the [=context object=].
 
 To <dfn lt="get a stream for a Blob|getting a stream for a Blob">get a stream for a {{Blob}}</dfn>
 |blob|, run the following steps:
 
 1. Let |stream| be the result of [=construct a ReadableStream object|constructing=] a
    {{ReadableStream}} object.
-1. Return |stream| and run the following steps [=in parallel=]:
+1. Run the following steps [=in parallel=]:
   1. While not all bytes of |blob| have been read:
     1. Let |bytes| be the byte sequence that results from reading a [=chunk=] from |blob|.
     1. If a [=file read error=] occured while reading |bytes|, [$ReadableStream/error$]
        |stream| with a [=failure reason=] and abort these steps.
     1. [=ReadableStream/Enqueue=] |bytes| into |stream|.
 
-    Issue: We need to specify more concretely what reading from a Blob actually does, and what
-    possible errors can happen.
+    Issue: We need to specify more concretely what reading from a Blob actually does,
+    and what possible errors can happen.
+1. Return |stream|.
 
-<h4 id="text-method-algo">
-The text method</h4>
+### The {{Blob/text()}} method ### {#text-method-algo}
 
-The <dfn method for=Blob>text()</dfn> method returns the contents
-of the blob as a {{USVString}}.
-
-It must act as follows:
+The <dfn method for=Blob>text()</dfn> method, when invoked, must run these steps:
 
 1. Let |stream| be the the result of [=getting a stream for a Blob=] for the [=context object=].
-1. Let |reader| be the result of [=getting a reader=] from |stream|.
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
    If that threw an exception, return a new promise rejected with that exception.
-1. Let |promise| be the result of [=reading all bytes=] from |stream| with |reader|.
+1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
 1. Return the result of transforming |promise| by a fulfillment handler that returns the result of
    running [=UTF-8 decode=] on its first argument.
 
-<h4 id="arraybuffer-method-algo">
-The arrayBuffer method</h4>
+### The {{Blob/arrayBuffer()}} method ### {#arraybuffer-method-algo}
 
-The <dfn method for=Blob>arrayBuffer()</dfn> method returns the contents
-of the blob as an {{ArrayBuffer}}.
-
-It must act as follows:
+The <dfn method for=Blob>arrayBuffer()</dfn> method, when invoked, must run these steps:
 
 1. Let |stream| be the the result of [=getting a stream for a Blob=] for the [=context object=].
-1. Let |reader| be the result of [=getting a reader=] from |stream|.
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
    If that threw an exception, return a new promise rejected with that exception.
-1. Let |promise| be the result of [=reading all bytes=] from |stream| with |reader|.
+1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
 1. Return the result of transforming |promise| by a fulfillment handler that returns
    a new {{ArrayBuffer} whose contents are its first argument.
 
--- a/index.bs
+++ b/index.bs
@@ -65,6 +65,9 @@ spec: url
     text: url; for:/
   type: interface
     text: URL
+spec: fetch
+  type:interface
+    text:ReadableStream
 </pre>
 
 <pre class="anchors">
@@ -215,7 +218,7 @@ which must be initially set to the state of the underlying storage,
 if any such underlying storage exists.
 Further normative definition of <a>snapshot state</a> can be found for {{File}}s.
 
-<pre class="idl">
+<xmp class="idl">
 [Constructor(optional sequence&lt;BlobPart> blobParts,
              optional BlobPropertyBag options),
  Exposed=(Window,Worker), Serializable]
@@ -228,6 +231,11 @@ interface Blob {
   Blob slice(optional [Clamp] long long start,
             optional [Clamp] long long end,
             optional DOMString contentType);
+
+  // read from the Blob.
+  ReadableStream stream();
+  Promise<USVString> text();
+  Promise<ArrayBuffer> arrayBuffer();
 };
 
 enum EndingType { "transparent", "native" };
@@ -238,7 +246,7 @@ dictionary BlobPropertyBag {
 };
 
 typedef (BufferSource or Blob or USVString) BlobPart;
-</pre>
+</xmp>
 
 {{Blob}} objects are [=serializable objects=]. Their [=serialization steps=],
 given |value| and |serialized|, are:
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
@annevk would you mind taking a look at this?

index.bs
```diff
@@ -551,6 +559,59 @@ It must act as follows:
   </pre>
 </div>
 
+<h4 id="stream-method-algo">
+The stream method</h4>
```
annevk: I guess this is technically up to you, but I'd prefer not wrapping after an opening tag or before a closing tag as it introduces whitespace where there should be none.

annevk: Also, if these are to have headings we should use `<code>stream()</code>` or some such as that's the typical way of writing down a method name.

index.bs
```diff
@@ -551,6 +559,59 @@ It must act as follows:
...
+
+The <dfn method for=Blob>stream()</dfn> method returns the contents
+of the blob as a {{ReadableStream}}.
```
annevk: Do we need this summary? It seems clearer if the `<dfn>` is immediately about the normative aspect.

index.bs
```diff
@@ -551,6 +559,59 @@ It must act as follows:
...
+
+It must return the result of [=getting a stream for a Blob=] for the [=context object=].
+
+To <dfn lt="get a stream for a Blob|getting a stream for a Blob">get a stream for a {{Blob}}</dfn>
+|blob|, run the following steps:
+
+1. Let |stream| be the result of [=construct a ReadableStream object|constructing=] a
+   {{ReadableStream}} object.
+1. Return |stream| and run the following steps [=in parallel=]:
```
annevk: Please do the return as the last step. So "2. Run these steps in parallel: 2.1 ... 3. Return _stream_."

index.bs
```diff
@@ -551,6 +559,59 @@ It must act as follows:
...
+  1. While not all bytes of |blob| have been read:
+    1. Let |bytes| be the byte sequence that results from reading a [=chunk=] from |blob|.
+    1. If a [=file read error=] occured while reading |bytes|, [$ReadableStream/error$]
+       |stream| with a [=failure reason=] and abort these steps.
+    1. [=ReadableStream/Enqueue=] |bytes| into |stream|.
+
+    Issue: We need to specify more concretely what reading from a Blob actually does, and what
+    possible errors can happen.
```
annevk: Yeah, and how large is a chunk? Is there some POSIX operation we can hint at?

index.bs
```diff
@@ -551,6 +559,59 @@ It must act as follows:
...
+
+<h4 id="text-method-algo">
+The text method</h4>
+
+The <dfn method for=Blob>text()</dfn> method returns the contents
+of the blob as a {{USVString}}.
+
+It must act as follows:
+
+1. Let |stream| be the the result of [=getting a stream for a Blob=] for the [=context object=].
```
annevk: So if we're going to define these in terms of streaming. Should we perhaps test this approach on `FileReader` first? So that the underlying streaming primitive is powerful enough to explain all these various features. For this particular method we could then likely use a shared abstraction as `FileReader` also has a string-returning feature.

index.bs
mkruisselbrink: I was mostly trying to be consistent with the rest of the spec, but yeah agree that this should be done differently. I've cleaned up the rest of the spec and rebased this on top of that.

index.bs
mkruisselbrink: Good point. This sentence really doesn't add much, and if we want any kind of non-normative bits those should be domintro notes. Removed it here and for the other methods.

index.bs

index.bs
mkruisselbrink: Not sure if size of chunks is ever going to really matter, as that shouldn't really be observable anyway... These new methods only return the whole stream in one go anyway, and while the existing FileReader methods do give you progress events those are more time-limited (with the same ~50ms as used by XHR) rather than directly tied to chunks that are pushed in the stream. And I expect this always to be somewhat hand-wavy, but at least I hope to have one place that describes how to get a ReadableStream out of a blob, and then the rest of the spec can just use that stream and not have any hand-waving or ambiguities... This "algorithm" at least doesn't seem worse than the current defintion of read operations.

index.bs
mkruisselbrink: Sure, I don't particularly mind doing things in either order. This seemed like a smaller change to start with, but the opposite is of course true as well. I'm working on the more general FileReader rewrite, and should have that uploaded soon too.

This also brings up the question if `blob.text()` should do what is described here, matching `new Response(blob).text()` (and always using UTF-8), or if it should do what `readAsText` does currently, and use the encoding that might be specified in the blob's mime type (or even have an optional encoding parameter to pass to the text method to override that default). Not sure which one is more important to be consistent with...

domenic: > This also brings up the question if blob.text() should

Oooh, interesting question. I tend to think it'd be a bit frustrating for authors and users if the charset was ignored, especially if interoperating with legacy systems or files on disk that are not properly UTF-8 encoded. But, I'm not sure.

mkruisselbrink: Any idea why `Body.text()` ignores the charset? It seems like it could be equally frustrating there, and the algorithm there does have access to the mime type as well.

annevk: I did it because it's a convenience method and everyone should migrate to UTF-8.

It's not clear to me how useful it would be to consider other encodings here if you only look at the MIME type as file systems wouldn't give you that information at that level, presumably.

index.bs
annevk: Well, the `stream()` method gives you incremental access, right? I was interested in that.

mkruisselbrink: Good point, yeah. Not sure what we can helpfully say about chunk sizes though... What does fetch do for example? Especially when a response is read from the http cache it seems like that is sort of a similar situation as this one? Does it specify anything about how the body is streamed?

index.bs
mkruisselbrink: https://github.com/w3c/FileAPI/pull/118 is my attempt at rewriting FileReader in similar terms as this pull request does. Let me know if you want me to finish landing that before getting back to this PR?

Just sticking to UTF-8 seems fine to me. At least it would be consistent with fetch.

mkruisselbrink:
(of course this will need tests before landing as well)

index.bs
annevk: It uses https://fetch.spec.whatwg.org/#concept-read-chunk-from-readablestream. As I mentioned in the other issue though there are some problems with how the abstraction works.

index.bs
annevk: I'd prefer that, as it's the right way to approach these kind of problems. Ensure the new primitive can be used to implement all the things, then expose fancy new APIs.

index.bs
mkruisselbrink: No, that is for reading chunks from the stream. The bit I can't find is how it is specified what chunks you're going to get out of a stream. I.e. if you call `(await fetch('....')).body`, and read from that. I don't think it is specified what those chunks are either? I can imagine if the fetch hit the network it would depend on what the server returns, but if it is loaded from the cache it is less clear. Fetch just seems to link to https://tools.ietf.org/html/rfc7234#section-4 for how the Response (and its body) are created when loaded from the cache, but also doesn't mention anything about chunk size etc.

mkruisselbrink:
Okay, rebased this on top of the FileReader changes.

index.bs
annevk: Ah sorry, see step 13 of https://fetch.spec.whatwg.org/#http-network-fetch. It basically waits for a number of bytes to arrive and then enqueues a Uint8Array. That's probably what you want to mimic to some extent, until there's better infra.

index.bs
```diff
@@ -228,6 +231,11 @@ interface Blob {
   Blob slice(optional [Clamp] long long start,
             optional [Clamp] long long end,
             optional DOMString contentType);
+
+  // read from the Blob.
+  ReadableStream stream();
+  Promise<USVString> text();
+  Promise<ArrayBuffer> arrayBuffer();
```
annevk: These can all be annotated with `[NewObject]`. Filed #121 for the remainder.

index.bs
```diff
@@ -561,6 +569,38 @@ It must act as follows:
...
+
+1. Let |stream| be the result of calling [=get stream=] on the [=context object=].
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
+   If that threw an exception, return a new promise rejected with that exception.
+1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
+1. Return the result of transforming |promise| by a fulfillment handler that returns the result of
+   running [=UTF-8 decode=] on its first argument.
+
+Note: This is different from the behavior of {{FileReader/readAsText()}} to align better
+with the behavior of {{Body/text()|Body.text()}}. Specifically this method will always
```
annevk: I'd make the link text something like "Fetch's text()" for clarity. Body is just a specification mechanism.

Jamesernator:
Any possibility of including `.dataURL()` as well?

annevk:
@Jamesernator let's track that separately.

mkruisselbrink:
Tests are in progress in https://chromium-review.googlesource.com/c/chromium/src/+/1526796.

Implementation bugs: [Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=945893), [WebKit](https://bugs.webkit.org/show_bug.cgi?id=196258), [FireFox](https://bugzilla.mozilla.org/show_bug.cgi?id=1539205)

jarryd999:
Hey @annevk, do you think Mozilla will implement this soon? I'm filing the Intent to Implement/Ship for blink and I'd like to accurately represent Mozilla's stance.

annevk:
I can't really comment on a timeline, but we're supportive of these methods being added to the web platform.

jimmywarting:
> Any possibility of including .dataURL() as well?

- I would suggest using URL.createObjectURL(blob) instead of using a base64 string
- if you want to upload/save a file you should send it as a binary

base64 is ~3x larger

jimmywarting:
Should the FileReader change how it gets the stream?

`Should Let stream be the result of calling get stream on blob .` ->
`Should Let stream be the result of calling blob.stream() .`

That would make fileReader a self contained module and it dosen't have to have any internal access to  blobs private methods

annevk:
@jimmywarting please file a new issue if you feel strongly, but generally we don't use that pattern to describe platform objects. If someone were to modify `Blob.prototype.stream` it shouldn't affect `FileReader`. It would actually add quite some complexity as we'd have to handle all the exceptional cases that might arise from such a setup.

jimmywarting:
hmm okey, will stay on the fence.

annevk:
@jimmywarting happy to discuss this further somewhere somehow. It kinda falls out of IDL and how browsers implement platform objects.  https://annevankesteren.nl/2015/01/javascript-web-platform explains some of this, but doesn't really go into why we try to avoid invoking JavaScript once we cross the IDL boundary.

Issue #79, opened 2017-05-04
Don't update FileReader's result until "LOAD"


annevk:
Otherwise you get a giant memory leak. Reportedly Firefox already does this. Chrome allocates strings/buffers and will quickly get slow.

jakearchibald:
Chrome bug https://bugs.chromium.org/p/chromium/issues/detail?id=674903

jakearchibald:
We could provide incremental updates for arraybuffers by allocating `new ArrayBuffer(file.size)` as soon as `readAsArrayBuffer` is called, then update the buffer for each progress event. This would avoid the performance issues Chrome is hitting.

Unfortunately it means that `reader.result.byteLength` will not reflect progress, which may be considered a backwards incompatible change.

jakearchibald:
However, XHR doesn't offer an arraybuffer until the request is complete, so aligning with that might be better. Then, create a streams-based API to do it properly.

Again, this may be backwards incompatible.

mkruisselbrink:
@jakearchibald that chrome issue was actually for a completely different bug (until you added comments about this issue to it). I do agree that it might make sense to get rid of result during progress events if that is what other browsers are already doing anyway. As you mention, should be relatively easy to add a use-counter for.

mkruisselbrink:
Actually, the spec already sort of implies that result shouldn't be available until LOAD time. All the various read method algorithms don't assign any value to `result` until the "process read EOF" steps. It's probably still a good idea to write some tests to figure out what browsers are actually doing here though (for example does `result` keep the result of a previous read operation until a new load has finished, or is it reset back to null when the new read starts?)

And I added the use counter to chrome to see how often this (non-spec-compliant) behavior of returning result before load is finished is actually triggered.

mkruisselbrink:
Chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=768972

mkruisselbrink:
And some tests to figure out the current behavior in https://github.com/w3c/web-platform-tests/pull/7494

In summary, currently Firefox seems to be the odd one out, with Chrome, Safari and Edge all having result available during progress events (at least during the last progress event; it's hard to come up with a cross-browser test that triggers multiple progress events). Firefox on the other hand seems to more closely follow the spec, with two exceptions: 1) when starting a new read it resets result to null, and 2) for readAsBinaryString Firefox actually does match the other three browsers and the result is available during the progress event. It's only for the other three read methods were it doesn't.

[...]

PR #118, opened 2019-03-12
Rewrite FileReader definitions.

mkruisselbrink:
Make algorithms more imperative to make it clearer what is going on.
No intended behavioral changes, but making it clearer that the behavior
from #79 is the expected behavior.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/pull/118.html" title="Last updated on Mar 14, 2019, 10:43 PM UTC (c8bb5a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/118/5688560...c8bb5a2.html" title="Last updated on Mar 14, 2019, 10:43 PM UTC (c8bb5a2)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -916,7 +916,7 @@ runs the following steps:
      [=queue a task=] to run the following steps and abort this algorithm:
     1. Set the [=context object=]'s {{FileReader/readyState}} to {{DONE}}.
     1. Let |result| be the result of
-       [=package data=] given |bytes|, |type|, |blob|'s {{Blob/type}} and |encodingName|.
+       [=package data=] given |bytes|, |type|, |blob|'s {{Blob/type}}, and |encodingName|.
     1. If [=package data=] threw an exception |error|:
       1. Set the [=context object=]'s {{FileReader/error!!attribute}} attribute to |error|.
       1. [=Fire a progress event=] called {{error!!event}} at the [=context object=].
@@ -1070,7 +1070,7 @@ the user agent must run the steps below:
 
 <div algorithm="package data">
 A {{Blob}} has an associated <dfn for=Blob>package data</dfn> algorithm,
-given |bytes|, a |type|, a optional |mimeType| and a optional |encodingName|,
+given |bytes|, a |type|, a optional |mimeType|, and a optional |encodingName|,
 switches on |type|, and runs the associated steps:
 
 : DataURL
@@ -1079,18 +1079,17 @@ switches on |type|, and runs the associated steps:
   * Use |mimeType| as part of the Data URL if it is available
     in keeping with the Data URL specification [[!RFC2397]].
   * If |mimeType| is not available return a Data URL without a media-type. [[!RFC2397]].
-    Data URLs that do not have media-types [[RFC2046]]
-    must be treated as plain text by conforming user agents. [[!RFC2397]].
+
+  Issue(104): Better specify how the DataURL is generated.
 
 : Text
-:: 1. Let |encoding| be null.
+:: 1. Let |encoding| be failure.
    1. If the |encodingName| is present, set |encoding| to the result of
       [=getting an encoding=] from |encodingName|.
-   1. If the [=getting an encoding=] step above returns failure, set |encoding| to null.
-   1. If |encoding| is null, and |mimeType| is present:
+   1. If |encoding| is failure, and |mimeType| is present:
      1. Let |type| be the result of [=parse a MIME type=] given |mimeType|.
-     1. If |type| is failure, abort these substeps.
-     1. Set |encoding| to the result of [=getting an encoding=]
+     1. If |type| is not failure,
+        set |encoding| to the result of [=getting an encoding=]
         from |type|'s [=MIME type/parameters=][`"charset"`].
 
       <div class="example">
@@ -1099,9 +1098,7 @@ switches on |type|, and runs the associated steps:
         Note that user agents must parse and extract the portion of the Charset Parameter
         that constitutes a *label* of an encoding.
       </div>
-     1. If the [=getting an encoding=] steps above return failure,
-        then set |encoding| to null.
-   1. If |encoding| is null, then set |encoding| to [=UTF-8=].
+   1. If |encoding| is failure, then set |encoding| to [=UTF-8=].
    1. [=Decode=] |bytes| using fallback encoding |encoding|, and return the result.
 
 : ArrayBuffer
@@ -1244,13 +1241,11 @@ when invoked, must run these steps:
 
 1. Let |stream| be the result of calling [=get stream=] on |blob|.
 1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
-   If that threw an exception, abort these steps and rethrow that exception.
 1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
 1. Wait for |promise| to be fulfilled or rejected.
 1. If |promise| fulfilled with a [=byte sequence=] |bytes|:
-  1. Return the result of [=package data=] given |bytes|, *Text*, |blob|'s {{Blob/type}} and |encoding|.
-1. Otherwise:
-  1. Throw whatever |promise| was rejected with.
+  1. Return the result of [=package data=] given |bytes|, *Text*, |blob|'s {{Blob/type}}, and |encoding|.
+1. Throw |promise|'s rejection reason.
 
 #### The {{FileReaderSync/readAsDataURL()}} method #### {#readAsDataURLSync-section}
 
@@ -1259,13 +1254,11 @@ when invoked, must run these steps:
 
 1. Let |stream| be the result of calling [=get stream=] on |blob|.
 1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
-   If that threw an exception, abort these steps and rethrow that exception.
 1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
 1. Wait for |promise| to be fulfilled or rejected.
 1. If |promise| fulfilled with a [=byte sequence=] |bytes|:
-  1. Return the result of [=package data=] given |bytes|, *DataURL* and |blob|'s {{Blob/type}}.
-1. Otherwise:
-  1. Throw whatever |promise| was rejected with.
+  1. Return the result of [=package data=] given |bytes|, *DataURL*, and |blob|'s {{Blob/type}}.
+1. Throw |promise|'s rejection reason.
 
 #### The {{FileReaderSync/readAsArrayBuffer()}} method #### {#readAsArrayBufferSyncSection}
 
@@ -1274,13 +1267,11 @@ when invoked, must run these steps:
 
 1. Let |stream| be the result of calling [=get stream=] on |blob|.
 1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
-   If that threw an exception, abort these steps and rethrow that exception.
 1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
 1. Wait for |promise| to be fulfilled or rejected.
 1. If |promise| fulfilled with a [=byte sequence=] |bytes|:
-  1. Return the result of [=package data=] given |bytes|, *ArrayBuffer* and |blob|'s {{Blob/type}}.
-1. Otherwise:
-  1. Throw whatever |promise| was rejected with.
+  1. Return the result of [=package data=] given |bytes|, *ArrayBuffer*, and |blob|'s {{Blob/type}}.
+1. Throw |promise|'s rejection reason.
 
 #### The {{FileReaderSync/readAsBinaryString()}} method #### {#readAsBinaryStringSyncSection}
 
@@ -1289,13 +1280,11 @@ when invoked, must run these steps:
 
 1. Let |stream| be the result of calling [=get stream=] on |blob|.
 1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
-   If that threw an exception, abort these steps and rethrow that exception.
 1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
 1. Wait for |promise| to be fulfilled or rejected.
 1. If |promise| fulfilled with a [=byte sequence=] |bytes|:
-  1. Return the result of [=package data=] given |bytes|, *BinaryString* and |blob|'s {{Blob/type}}.
-1. Otherwise:
-  1. Throw whatever |promise| was rejected with.
+  1. Return the result of [=package data=] given |bytes|, *BinaryString*, and |blob|'s {{Blob/type}}.
+1. Throw |promise|'s rejection reason.
 
 Note: The use of {{FileReaderSync/readAsArrayBuffer()}} is preferred over
 {{FileReaderSync/readAsBinaryString()}}, which is provided for
--- a/index.bs
+++ b/index.bs
@@ -261,7 +261,7 @@ which runs these steps:
    {{ReadableStream}} object.
 1. Run the following steps [=in parallel=]:
   1. While not all bytes of |blob| have been read:
-    1. Let |bytes| be the byte sequence that results from reading a [=chunk=] from |blob|.
+    1. Let |bytes| be the [=byte sequence=] that results from reading a [=chunk=] from |blob|.
     1. If a [=file read error=] occured while reading |bytes|, [$ReadableStream/error$]
        |stream| with a [=failure reason=] and abort these steps.
     1. [=ReadableStream/Enqueue=] |bytes| into |stream|.
@@ -888,7 +888,7 @@ interface FileReader: EventTarget {
 
 <div algorithm="read operation">
 A {{FileReader}} has an associated <dfn id=readOperation>read operation</dfn> algorithm,
-given |blob|, a |type| and an optional |encodingName|,
+which given |blob|, a |type| and an optional |encodingName|,
 runs the following steps:
 
 1. Set the [=context object=]'s {{FileReader/result}} to `null`.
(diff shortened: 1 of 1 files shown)


mkruisselbrink:
This isn't supposed to change behavior, but I should still write tests to actually verify the behavior, as there are things in this area (like #79) that aren't currently tested, and where browsers do diverge.

index.bs
```diff
@@ -251,6 +253,25 @@ Their [=deserialization step=], given |serialized| and |value|, are:
 
 2. Set |value|'s underlying byte sequence to |serialized|.\[[ByteSequence]].
 
+<div algorithm="get stream">
+A {{Blob}} has an associated <dfn for=Blob>get stream</dfn> algorithm,
+given a |blob|, and runs the following steps:
```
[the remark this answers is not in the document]
mkruisselbrink: Ah yes, it doesn't make sense to both be associated with a `Blob` and take one. Changed it to be just associated with one.

index.bs
```diff
@@ -954,17 +883,72 @@ interface FileReader: EventTarget {
...
+runs the following steps:
+
+1. Set the [=context object=]'s {{FileReader/result}} to `null`.
+1. Set the [=context object=]'s {{FileReader/error!!attribute}} to `null`.
+1. Let |stream| be the result of calling [=get stream=] on |blob|.
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
+1. Let |bytes| by an empty [=byte sequence=].
+1. Let |chunk| be the result of [=read a chunk|reading a chunk=] from |stream| with |reader|.
+1. Let |isFirstChunk| be true.
+1. [=In parallel=], while true:
```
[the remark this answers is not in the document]
mkruisselbrink: This is exactly what fetch does in for example the ["transmit body"](https://fetch.spec.whatwg.org/#ref-for-concept-read-chunk-from-readablestream) algorithm. But yeah, not sure what the best spec language is these days to "wait" for promises to resolve without blocking the main thread of execution. Open to suggestions.

index.bs
```diff
@@ -954,17 +883,72 @@ interface FileReader: EventTarget {
...
+  1. Wait for |chunk| to be fulfilled or rejected.
+  1. If |chunk| is fulfilled, and |isFirstChunk| is true,
+     [=queue a task=] to [=fire a progress event=] called {{loadstart}} at the [=context object=].
+  1. Set |isFirstChunk| to false.
+
+  1. If |chunk| is fulfilled with an object whose `done` property is false and whose `value`
+     property is a `Uint8Array` object, then run these steps:
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -954,17 +883,72 @@ interface FileReader: EventTarget {
...
+    1. Let |bs| be the [=byte sequence=] represented by the `Uint8Array` object.
+    1. Append |bs| to |bytes|.
+    1. If roughly 50ms have passed since these subsubsubsteps were last invoked,
```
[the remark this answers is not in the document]
mkruisselbrink: Done (XHR seems to really like using subsubsteps and subsubsubsteps in its algorithsm, but I argee that it doesn't add much).

index.bs
```diff
@@ -1000,272 +984,135 @@ that user agents must support on {{FileReader}} as DOM attributes:
...
+    Data URLs that do not have media-types [[RFC2046]]
+    must be treated as plain text by conforming user agents. [[!RFC2397]].
+
+: Text
+:: 1. Let |encoding| be null.
+   1. If the |encodingName| is present, set |encoding| to the result of
+      [=getting an encoding=] from |encodingName|.
+   1. If the [=getting an encoding=] step above returns failure, set |encoding| to null.
+   1. If |encoding| is null, and |mimeType| is present,
+      and it uses a Charset Parameter [[RFC2046]],
```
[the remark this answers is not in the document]

index.bs
```diff
@@ -1000,272 +984,135 @@ that user agents must support on {{FileReader}} as DOM attributes:
...
+        then <a>getting an encoding</a> is run using `"utf-8"` as the label.
+        Note that user agents must parse and extract the portion of the Charset Parameter
+        that constitutes a *label* of an encoding.
+      </div>
+   1. If the [=getting an encoding=] steps above return failure,
+      then set |encoding| to null.
+   1. If |encoding| is null, then set |encoding| to [=UTF-8=].
+   1. [=Decode=] |bytes| using fallback encoding |encoding|, and return the result.
+
+   The above steps are also known as the <dfn>encoding determination</dfn>.
```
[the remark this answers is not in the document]
mkruisselbrink: Not sure what you mean with this? But removed the definition anyway, as it does indeed not really serve a purpose anymore.

mkruisselbrink:
Thanks for the review. Yeah, I tried to mostly just mimic what fetch (and sometimes XHR) are doing.

Corresponding tests are in https://github.com/web-platform-tests/wpt/pull/7494 (and unfortunately not a single browser currently passes those tests... Firefox is close (it is synchronously firing the loadstart event rather than posting a task, and it set result too soon for readAsBinaryString). Chromium/Edge have various other issues, but we already have a bug for at least some of those, and I don't see any problems with fixing those. Safari seems to mostly currently match Chromium in behavior.

index.bs
```diff
@@ -1000,272 +984,134 @@ that user agents must support on {{FileReader}} as DOM attributes:
...
+    in keeping with the Data URL specification [[!RFC2397]].
+  * If |mimeType| is not available return a Data URL without a media-type. [[!RFC2397]].
+    Data URLs that do not have media-types [[RFC2046]]
+    must be treated as plain text by conforming user agents. [[!RFC2397]].
+
+: Text
+:: 1. Let |encoding| be null.
+   1. If the |encodingName| is present, set |encoding| to the result of
+      [=getting an encoding=] from |encodingName|.
+   1. If the [=getting an encoding=] step above returns failure, set |encoding| to null.
```
annevk: It seems you only need to check if encoding is failure at this point. That's cleaner than referencing an algorithm in a prior step.

index.bs
```diff
@@ -1000,272 +984,134 @@ that user agents must support on {{FileReader}} as DOM attributes:
...
+   1. If |encoding| is null, and |mimeType| is present:
+     1. Let |type| be the result of [=parse a MIME type=] given |mimeType|.
+     1. If |type| is failure, abort these substeps.
```
annevk: Might be good to file a follow-up bug on removing this kind of anti-pattern. A preferred style would be checking that type is not failure and then nesting the further set of steps, or something equivalent.

index.bs
```diff
@@ -1391,71 +1237,69 @@ interface FileReaderSync {
...
-  and with the <a>synchronous flag</a> *set*.
-  If the read operation returns failure,
-  throw the appropriate exception as defined in [[#dfn-error-codes]].
-  <a>Terminate this algorithm</a>.
-1. If no error has occurred,
-  return the |result| of the <a>read operation</a> represented as a string
-  in a format determined through the <a>encoding determination</a> algorithm.
+1. Let |stream| be the result of calling [=get stream=] on |blob|.
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
+   If that threw an exception, abort these steps and rethrow that exception.
```
annevk: This is implied these days so no need to state it.

index.bs
```diff
@@ -1391,71 +1237,69 @@ interface FileReaderSync {
...
+1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
+1. Wait for |promise| to be fulfilled or rejected.
+1. If |promise| fulfilled with a [=byte sequence=] |bytes|:
+  1. Return the result of [=package data=] given |bytes|, *Text*, |blob|'s {{Blob/type}} and |encoding|.
```
annevk: Oxford comma

index.bs
```diff
@@ -1391,71 +1237,69 @@ interface FileReaderSync {
...
+1. Otherwise:
```
annevk: No need for otherwise here as it can only happen if the above didn't happen, right?

index.bs
```diff
@@ -1391,71 +1237,69 @@ interface FileReaderSync {
...
-1. If no error has occurred, return the |result| of the <a>read operation</a>
-  as a Data URL [[!RFC2397]] subject to the considerations below:
-
-  * Use the <code>blob</code>'s {{Blob/type}} attribute as part of the Data URL if it is available
-    in keeping with the Data URL specification [[!RFC2397]].
-  * If the {{Blob/type}} attribute is not available on the <code>blob</code> return a Data URL without a media-type. [[!RFC2397]].
-    Data URLs that do not have media-types [[RFC2046]] must be treated as plain text by conforming user agents. [[!RFC2397]].
+1. Let |stream| be the result of calling [=get stream=] on |blob|.
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
+   If that threw an exception, abort these steps and rethrow that exception.
```
annevk: This can be removed per above.

index.bs
```diff
@@ -1391,71 +1237,69 @@ interface FileReaderSync {
...
 
-1. Initiate a <a>read operation</a> using the <code>blob</code> argument,
-  and with the <a>synchronous flag</a> *set*.
-  If the <a>read operation</a> returns failure,
-  throw the appropriate exception as defined in [[#dfn-error-codes]].
-  <a>Terminate this algorithm</a>.
-1. If no error has occurred, return the |result| of the <a>read operation</a> as an {{ArrayBuffer}}.
+1. Let |stream| be the result of calling [=get stream=] on |blob|.
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
+   If that threw an exception, abort these steps and rethrow that exception.
```
annevk: Same.

index.bs
```diff
@@ -1391,71 +1237,69 @@ interface FileReaderSync {
...
+1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
+1. Wait for |promise| to be fulfilled or rejected.
+1. If |promise| fulfilled with a [=byte sequence=] |bytes|:
+  1. Return the result of [=package data=] given |bytes|, *ArrayBuffer* and |blob|'s {{Blob/type}}.
```
annevk: Oxford comma.

index.bs
```diff
@@ -954,17 +883,72 @@ interface FileReader: EventTarget {
...
+1. Set the [=context object=]'s {{FileReader/error!!attribute}} to `null`.
+1. Let |stream| be the result of calling [=get stream=] on |blob|.
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
+1. Let |bytes| by an empty [=byte sequence=].
+1. Let |chunk| be the result of [=read a chunk|reading a chunk=] from |stream| with |reader|.
+1. Let |isFirstChunk| be true.
+1. [=In parallel=], while true:
+  1. Wait for |chunk| to be fulfilled or rejected.
+  1. If |chunk| is fulfilled, and |isFirstChunk| is true,
+     [=queue a task=] to [=fire a progress event=] called {{loadstart}} at the [=context object=].
```
annevk: So I'd argue Firefox is correct in firing this synchronously. That's also what XHR does iirc. Is there a problem with doing that?

annevk: I think my only argument is consistency with the initial API that had `loadstart`. (And perhaps poor initial review by Mozilla.)

annevk: Filed #119.

index.bs
```diff
@@ -251,6 +253,25 @@ Their [=deserialization step=], given |serialized| and |value|, are:
...
+A {{Blob}} |blob| has an associated <dfn for=Blob>get stream</dfn> algorithm,
+which runs these steps:
+
+1. Let |stream| be the result of [=construct a ReadableStream object|constructing=] a
+   {{ReadableStream}} object.
+1. Run the following steps [=in parallel=]:
+  1. While not all bytes of |blob| have been read:
+    1. Let |bytes| be the byte sequence that results from reading a [=chunk=] from |blob|.
+    1. If a [=file read error=] occured while reading |bytes|, [$ReadableStream/error$]
+       |stream| with a [=failure reason=] and abort these steps.
```
domenic: Probably scope-creep to fix this here, but it's not very clear how to get from "failure reason" to `an "X" DOMException`.

index.bs
```diff
@@ -251,6 +253,25 @@ Their [=deserialization step=], given |serialized| and |value|, are:
...
 
+<div algorithm="get stream">
+A {{Blob}} |blob| has an associated <dfn for=Blob>get stream</dfn> algorithm,
+which runs these steps:
+
+1. Let |stream| be the result of [=construct a ReadableStream object|constructing=] a
+   {{ReadableStream}} object.
+1. Run the following steps [=in parallel=]:
+  1. While not all bytes of |blob| have been read:
+    1. Let |bytes| be the byte sequence that results from reading a [=chunk=] from |blob|.
```
domenic:
```suggestion
    1. Let |bytes| be the [=byte sequence=] that results from reading a [=chunk=] from |blob|.
```

index.bs
```diff
@@ -251,6 +253,25 @@ Their [=deserialization step=], given |serialized| and |value|, are:
...
+    1. If a [=file read error=] occured while reading |bytes|, [$ReadableStream/error$]
+       |stream| with a [=failure reason=] and abort these steps.
+    1. [=ReadableStream/Enqueue=] |bytes| into |stream|.
```
domenic: Readable streams, at least currently, take JS objects, not Infra things like byte sequences. For example https://fetch.spec.whatwg.org/#concept-read-all-bytes-from-readablestream assumes that the chunks are Uint8Arrays. So, if the goal is to integrate with that, you'd want to explicitly construct a Uint8Array.

Alternately, you could leave this as byte sequences as that's arguably a more correct long-term infrastructure choice. It's just the way that readable streams are defined right now that's making things difficult. I'm not sure what's best.

index.bs
```diff
@@ -954,17 +883,72 @@ interface FileReader: EventTarget {
   attribute EventHandler onabort;
   attribute EventHandler onerror;
   attribute EventHandler onloadend;
-
 };
 </pre>
 
+<div algorithm="read operation">
+A {{FileReader}} has an associated <dfn id=readOperation>read operation</dfn> algorithm,
+given |blob|, a |type| and an optional |encodingName|,
```
domenic:
```suggestion
which given |blob|, a |type| and an optional |encodingName|,
```

index.bs
```diff
@@ -954,17 +883,72 @@ interface FileReader: EventTarget {
...
+runs the following steps:
+
+1. Set the [=context object=]'s {{FileReader/result}} to `null`.
```
domenic: Maybe not something to clean up now, but I'd rather say "A FileReader _fr_ has an associated..." and then use _fr_ instead of the context object throughout.

index.bs
```diff
@@ -954,17 +883,72 @@ interface FileReader: EventTarget {
...
+1. Set the [=context object=]'s {{FileReader/error!!attribute}} to `null`.
```
domenic: Setting readonly attributes is not great; this should set an internal slot/associated value which those getters return the value of.

index.bs
```diff
@@ -954,17 +883,72 @@ interface FileReader: EventTarget {
...
+1. Let |stream| be the result of calling [=get stream=] on |blob|.
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
+1. Let |bytes| by an empty [=byte sequence=].
+1. Let |chunk| be the result of [=read a chunk|reading a chunk=] from |stream| with |reader|.
```
domenic: It's confusing to me to use _chunk_ to refer to a promise; I'd prefer _chunkPromise_.

index.bs
```diff
@@ -1000,272 +984,131 @@ that user agents must support on {{FileReader}} as DOM attributes:
...
-  <dd>The entire {{File}} or {{Blob}} has been read into memory,
-    OR a <a>file read error</a> occurred,
-    OR the read was aborted using {{FileReader/abort()}}.
-    The {{FileReader}} is no longer reading a {{File}} or {{Blob}}.
-    If {{FileReader/readyState}} is set to {{FileReader/DONE}}
-    it means at least one of the <a>read methods</a> have been called on this {{FileReader}}.
-</dl>
+The {{FileReader/readyState}} attribute tells you in which state the object is:
+
+: {{EMPTY}} (numeric value 0)
```
domenic: Personally I like having the dfns here and the IDL link to them. For constants it could go either way, but right now the readyState getter is not defined anywhere. (Related to the internal slots thing.)

index.bs
```diff
@@ -1000,272 +984,131 @@ that user agents must support on {{FileReader}} as DOM attributes:
...
+1. [=Fire a progress event=] called {{abort}} at the [=context object=].
+1. If [=context object=]'s {{FileReader/readyState}} is not {{LOADING}},
+   [=fire a progress event=] called {{loadend}} at the [=context object=].
+
+## Packaging data ## {#packaging-data}
+
+<div algorithm="package data">
+A {{Blob}} has an associated <dfn for=Blob>package data</dfn> algorithm,
+given |bytes|, a |type|, a optional |mimeType|, and a optional |encodingName|,
+switches on |type|, and runs the associated steps:
```
domenic:
```suggestion
which switches on |type| and runs the associated steps:
```

index.bs
```diff
@@ -1391,71 +1234,61 @@ interface FileReaderSync {
...
-  If the read operation returns failure,
-  throw the appropriate exception as defined in [[#dfn-error-codes]].
-  <a>Terminate this algorithm</a>.
-1. If no error has occurred,
-  return the |result| of the <a>read operation</a> represented as a string
-  in a format determined through the <a>encoding determination</a> algorithm.
+1. Let |stream| be the result of calling [=get stream=] on |blob|.
+1. Let |reader| be the result of [=get a reader|getting a reader=] from |stream|.
+1. Let |promise| be the result of [=read all bytes|reading all bytes=] from |stream| with |reader|.
+1. Wait for |promise| to be fulfilled or rejected.
```
domenic: Not the most sound thing to do with real promises, but let's pretend these are spec promises, like they really should be if streams had a better infrastructure :).

index.bs
mkruisselbrink: Done (and in doing so, realized that I somehow had lost the step that sets readyState to loading. Fixed that as well).

PR #125, opened 2019-04-05
Modernize Blob constructor a tiny bit

### LINKED ISSUE #54: Whatdoes it mean to invoke the Blob constructor with zero parameters?
For example, what should this do:

   var b = new Blob(undefined, { type: "foo" });

?  Should `b.type` be empty string or `"foo"`?  In any case, there is no concept of "number of parameters" for IDL implementation algorithms, so please define the behavior in terms of concepts that do exist.

annevk:
All implementations do something different for new Blob(undefined, { type: "x/x" }). Let's do something simple.

Fixes #54.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/pull/125.html" title="Last updated on May 16, 2019, 6:58 AM UTC (abb6af3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/125/9977ce8...abb6af3.html" title="Last updated on May 16, 2019, 6:58 AM UTC (abb6af3)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -216,7 +216,7 @@ if any such underlying storage exists.
 Further normative definition of <a>snapshot state</a> can be found for {{File}}s.
 
 <pre class="idl">
-[Constructor(optional sequence&lt;BlobPart> blobParts,
+[Constructor(optional sequence&lt;BlobPart> blobParts = [],
              optional BlobPropertyBag options),
  Exposed=(Window,Worker), Serializable]
 interface Blob {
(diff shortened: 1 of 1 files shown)


marcoscaceres:
need to close and reopen to trigger new travis

marcoscaceres:
@annevk, could you please fix the merge conflicts? The constructor can now be updated to new syntax.

annevk:
Once there are sufficient tests I'm happy to address the remainder (including filing bugs).

marcoscaceres:
ok, no probs. Was worried about bit rot. How about just getting it to a good state and we just don't merge until there are tests? The only thing I was unsure about was the merge conflict with the algorithm, the IDL seems like a quick fix.

marcoscaceres:
@annevk, just saw this was still pending. Should this be updated and merged?

annevk:
@wisniewskit did you end up doing https://github.com/whatwg/mimesniff/issues/189#issuecomment-2090884819 in Gecko? Maybe I should take some time to make the specification say that to try to move this along.

saschanaz:
Looks like the plan is to change our behavior first and see how whether it's web compatible, before changing WPT: https://phabricator.services.mozilla.com/D208848

wisniewskit:
@annevk, I haven't yet landed the patches for Gecko to try this, as I'm still figuring out which test cases in our own codebase fail, and how to deal with them. But it's not looking intractable, so hopefully we can land it sometime soon and get data on whether there's any obvious fallout on sites. (Of course it would be faster and fairer if others also did the same, so I welcome anyone willing to help push other vendors along as well!)

PR #127, opened 2019-04-13
Fix escaping that doesn't work in <xmp>

foolip:
Broken by https://github.com/w3c/FileAPI/pull/117


mkruisselbrink:
Oops, thank you for the fix!

PR #151, opened 2020-04-06
Use valid UUID in blob: URL example

ctavan:
The algorithm for creating a blob URL references RFC4122 to generate a UUID. However the example contains an _invalid_ UUID (according to RFC4122): `9115d58c-bcda-ff47-86e5-083e9a215304`

This given string is not a valid UUID because it does not contain a meaningful version field (the first character after the second dash).

It could be worthwhile to consider specifying the specific UUID algorithm to be used ([Version 4 UUIDs](https://tools.ietf.org/html/rfc4122#section-4.4)) however that might be an actual change in the spec, not just an editorial change like the one I propose here.

***
<a href="https://pr-preview.s3.amazonaws.com/ctavan/FileAPI/pull/151.html" title="Last updated on Apr 6, 2020, 12:26 PM UTC (f52617a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/151/2183e01...ctavan:f52617a.html" title="Last updated on Apr 6, 2020, 12:26 PM UTC (f52617a)">Diff</a>


mkruisselbrink:
@marcoscaceres any idea what this "Travis CI - Branch" is this PR seems to be blocked on? I'd expect "Travis CI - Pull Request" to be required, not the branch version, since afaik the branch version doesn't work for pull requests made from forks. Is this some misconfiguration?

marcoscaceres:
Oh, this might be stuck because of the trivisci.com vs travisci.org change... I'll just merge.

PR #154, opened 2020-05-21
Rewrite a lot of the internals of Blobs to make reading more explicit.

### LINKED ISSUE #47: Define WebKit's underlying file changes behavior
See https://bugzilla.mozilla.org/show_bug.cgi?id=660148 for details.

### LINKED ISSUE #71: Describe internal slots of Blob and File objects better
That way other standards can be more precise. E.g., I need this to fix https://github.com/whatwg/xhr/issues/75.

### LINKED ISSUE #75: Disallow mutating state of Blob/File objects
In particular, I think we should no longer add something like `close()` and make that an invariant of these objects.

That way there's no longer a need for defensive cloning (through serializing/deserializing) of these objects, which helps with https://github.com/whatwg/url/issues/277.

Making a note of this somewhere in the document seems sufficient.

### LINKED ISSUE #99: Does the specification define input[type=file] .files expected behaviour to relevant to reflecting real-time changes to user selected file?
Consider the HTML and JavaScript  posted at this Stack Overflow question [Input file size and content do not update on macOS](https://stackoverflow.com/questions/50370493/input-file-size-and-content-do-not-update-on-macos)

HTML 

```
<!DOCTYPE html>
<input type="file" id="file" />
<p id="info"></p>
```

JavaScript

```
window.addEventListener('load', function() {
  window.setInterval(function() {
    var logFile = document.querySelector('#file').files[0];
    if (logFile) {
      document.querySelector('#info').innerHTML = '<br/>' +
        (new Date()).toString() + '<br/>Last modified: ' +
        logFile.lastModified +
        '<br/>Size: ' +
        logFile.size;
    }
  }, 1000);
});
```

Reproduce steps as described by OP of the question

> Please test with following snippet. Select a file (for example a text file), see last modified timestamp and file size, then change file and look again, if size has changed. On macOS the file size doesn't change.

Using a sample file, for example, https://gist.github.com/guest271314/11edc4566ba94f204dd46e6ae26edaad#file-d134213ba9465cb74dfd36cde47bf102638c4a9f3aca357f79ee747dd5f49f1e0f0de , which contains the text

`123`

**Behaviour**

Chromium 65 

Once the local file is selected by user action and the local file is changed at a text editor, for example, by typing `4` following the existing `123` at the file (for example, the above linked gist) both `lastModified` and `size` properties are reflected as being changed client side, indicating changes to local file are being updated in real time though the user has not performed the affirmative action of re-selecting the file.

Firefox Trunk (60)

Once the local file is selected by user action and the local file is changed at a text editor, for example, by typing `4` following the existing `123` at the file (for example, the above linked gist) both `lastModified` and `size` properties are **not** reflected as being changed client side, indicating changes to local file are being **not** updated in real time where the user has not performed the affirmative action of re-selecting the file.

What does the specification define as the expected behaviour of `size` and `lastModified` properties of a `File` object within a `FileList` object where changes to an underlying local file are made **_after_** user selection of a local file?

mkruisselbrink:
This redefines how blobs work by adding internal slots and hooks to
support Blob objects backed by various sources of data. By doing so this
tries to make more explicit how various edge cases behave, as well as
making it clear that blobs are immutable.

This fixes #75, fixes #99, and fixes #47.
It also lays the ground work to address w3c/webappsec-clear-site-date#49.

This also fixes #71.

Implementation commitment:

 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
 * [x] Chromium (matches the currently implemented behavior)
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)


mkruisselbrink:
This at least matches how chrome implements blobs backed by files, and how slicing and/or combining those blobs with other blobs works. I haven't verified yet if all browsers actually fully agree on this though.

There also is still plenty of room for improvement here, and of course follow-up work to have HTML actually call out to the new "create a file backed File object" to create files in drag&drop and upload algorithms.

@annevk do you have time to review this? not urgent, since I won't have much time in the near future to follow up anyway, but definitely would welcome your input about both the high-level idea of how this tries to define how things work, as well as details.

index.bs
```diff
@@ -269,88 +304,146 @@ which runs these steps:
...
-      Normative conditions for this member are provided in the [[#constructorBlob]].
-    * <dfn dict-member for="BlobPropertyBag">endings</dfn>,
-      an enum which can take the values {{"transparent"}} or {{"native"}}.
-      By default this is set to {{"transparent"}}. If set to {{"native"}},
-      [=convert line endings to native|line endings will be converted to native=]
-      in any {{USVString}} elements in {{blobParts}}.
-</dl>
+1. Let |read state| be a new [=bytes blob read state=].
+1. If |offset| is larger than |snapshot state|[<code><a for="bytes blob snapshot state">"data"</a></code>]'s [=byte sequence/length=],
+  set |read state|.[=bytes blob read state/bytes=] to an empty [=byte sequence=].
```
inexorabletash, line 408: FWIW, I don't see the _struct_._name_ syntax called out in Infra. Do you know if there's plan to add it?

annevk:
It would help a lot to have an enumeration of the edge cases this attempts to cover as well as some kind of overview of the design to solve them.

index.bs
```diff
@@ -657,132 +882,160 @@ dictionary FilePropertyBag : BlobPropertyBag {
...
+  derived from |name| in a user agent defined manner, or an empty string if no type
+  could be determined, taking into account the following <dfn export>file type guidelines</dfn>:
+
+  * User agents must return the {{Blob/type}} as an ASCII-encoded string in lower case,
+    such that when it is converted to a corresponding byte sequence,
+    it is a <a>parsable MIME type</a>,
+    or the empty string &ndash; 0 bytes &ndash; if the type cannot be determined.
+  * When the file is of type <code>text/plain</code>
+    user agents must NOT append a charset parameter to the <i>dictionary of parameters</i> portion of the media type [[!MIMESNIFF]].
+  * User agents must not attempt heuristic determination of encoding,
```
inexorabletash, line 932: Since encoding is not specified for `text/plain` (per previous), this can't apply in that case. Should this be scoped to other text/* types?

index.bs
```diff
@@ -657,132 +882,160 @@ dictionary FilePropertyBag : BlobPropertyBag {
...
-  1. If the {{BlobPropertyBag/type}} member is provided and is not the empty string,
-    let |t| be set to the {{BlobPropertyBag/type}} dictionary member.
-    If |t| contains any characters outside the range U+0020 to U+007E,
-    then set |t| to the empty string and return from these substeps.
-  2. Convert every character in |t| to [=ASCII lowercase=].
-  3. If the {{FilePropertyBag/lastModified}} member is provided,
-    let |d| be set to the  {{FilePropertyBag/lastModified}} dictionary member.
-    If it is not provided,
-    set |d| to the current date and time
+  with constructed files, mandating UTF-16 lessens ambiquity when file names are converted to <a>byte sequences</a>.
```
inexorabletash, line 996: spelling: should be "ambiguity"

index.bs
```diff
@@ -657,132 +882,160 @@ dictionary FilePropertyBag : BlobPropertyBag {
...
-1. Let |bytes| be the result of [=processing blob parts=] given {{fileBits}}
-  and {{File/File(fileBits, fileName, options)/options}}.
+The <dfn constructor for=File lt="File(fileBits, fileName, options)|File(fileBits, fileName)"><code>new File(|fileBits|, |fileName|, |options|)</code></dfn> constructor steps are:
 
+1. Run the {{Blob(blobParts, options)|Blob}}(|fileBits|, |options|) constructor steps.
 
-2. Let |n| be a new string of the same size as the {{fileName}} argument to the constructor.
-  Copy every character from {{fileName}} to |n|,
+1. Let |n| be a new string of the same size as |fileName|.
+1. Copy every character from |fileName| to |n|,
```
inexorabletash, line 992: Not new: use "code point" not "character" here and elsewhere.

index.bs
```diff
@@ -657,132 +882,160 @@ dictionary FilePropertyBag : BlobPropertyBag {
...
-    then set |t| to the empty string and return from these substeps.
-  2. Convert every character in |t| to [=ASCII lowercase=].
-  3. If the {{FilePropertyBag/lastModified}} member is provided,
-    let |d| be set to the  {{FilePropertyBag/lastModified}} dictionary member.
-    If it is not provided,
-    set |d| to the current date and time
+  with constructed files, mandating UTF-16 lessens ambiquity when file names are converted to <a>byte sequences</a>.
+
+1. If |options|.{{FilePropertyBag/lastModified}} member is provided:
+  1. Let |d| be |options|.{{FilePropertyBag/lastModified}} dictionary member.
```
inexorabletash, line 999: consistency: "member" or "dictionary member"

index.bs
```diff
@@ -657,132 +882,160 @@ dictionary FilePropertyBag : BlobPropertyBag {
...
+1. Otherwise:
+  1. Let |d| be the current date and time
```
inexorabletash, line 1001: Current date / time (as millis since epoch) comes up several times in the spec. Factor it out into a definition.

index.bs
```diff
@@ -657,132 +882,160 @@ dictionary FilePropertyBag : BlobPropertyBag {
...
     represented as the number of milliseconds since the <a>Unix Epoch</a>
     (which is the equivalent of <code>Date.now()</code> [[ECMA-262]]).
 
     Note: Since ECMA-262 {{Date}} objects convert to <code>long long</code> values
```
inexorabletash, line 1005: Maybe move this into the domintro ?

inexorabletash: I guess File constructor doesn't have a domintro yet. Add one? Blob too?

That'd be a good place to call out (in giant blinking text...) that the first argument is a array. :)

PR #155, opened 2020-06-18
Fix typo in read operation algorithm description

customcommander:


mkruisselbrink:
@marcoscaceres there still seems to be something messed up with the Travis configuration for this repo. It doesn't make sense for it to be blocked on "Travis CI - Branch" for pull requests that come from forks.

PR #160, opened 2021-01-22
Use GitHub workflows instead of travis.org for CI and publishing.

mkruisselbrink:


mkruisselbrink:
This doesn't currently do w3c publishing yet, since afaik the w3c spec-prod github action doesn't support that for bikeshed specs yet. Since we probably don't want to lose that, we should hold of on merging this...

annevk:
Maybe @sideshowbarker can help with this as he recently converted a bunch of repos.

.github/workflows/pr-push.yml
```diff
@@ -0,0 +1,16 @@
...
+  push:
+    branches: [main]
+
+jobs:
+  main:
+    name: Build, Validate and Deploy
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: sidvishnoi/spec-prod@v1
```
sideshowbarker, line 13:
```suggestion
      - uses: w3c/spec-prod@v2
```

sideshowbarker (changes requested):
With the suggested commit applied, this should work as-is

sideshowbarker (approved):

mkruisselbrink:
I think this is still missing the auto-publishing as WD? Or does that not need some configuration as well?

sideshowbarker:
> I think this is still missing the auto-publishing as WD? Or does that not need some configuration as well?

Yeah it needs a couple more lines in the config file. But I’ve now added them, in https://github.com/w3c/FileAPI/pull/167.

PR #164, opened 2021-04-19
Editorial: move from context object to this

annevk:
***
<a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/pull/164.html" title="Last updated on Apr 21, 2021, 6:38 AM UTC (5ba8fb9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/164/90d7bc7...5ba8fb9.html" title="Last updated on Apr 21, 2021, 6:38 AM UTC (5ba8fb9)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -66,9 +66,6 @@ spec: url
     text: url; for:/
   type: interface
     text: URL
-spec: fetch
-  type:interface
-    text:ReadableStream
 </pre>
 
 <pre class="anchors">
@@ -491,8 +488,7 @@ up to but not including the optional {{end}} parameter,
 and with a {{Blob/type}} attribute that is the value of the optional {{contentType!!argument}} parameter.
 It must act as follows:
 
-1. Let |O| be the {{Blob}} <a>context object</a> on which the {{slice()}} method is being called.
-2. The optional <dfn argument for="Blob/slice(start, end, contentType)" id="dfn-start">start</dfn> parameter
+1. The optional <dfn argument for="Blob/slice(start, end, contentType)" id="dfn-start">start</dfn> parameter
   is a value for the start point of a {{slice()}} call,
   and must be treated as a byte-order position,
   with the zeroth position representing the first byte.
@@ -504,7 +500,7 @@ It must act as follows:
     <li>Else, let |relativeStart| be <code>min(start, size)</code>.
   </ol>
 
-3. The optional <dfn argument for="Blob/slice(start, end, contentType)" id="dfn-end">end</dfn> parameter
+2. The optional <dfn argument for="Blob/slice(start, end, contentType)" id="dfn-end">end</dfn> parameter
   is a value for the end point of a {{slice()}} call.
   User agents must process {{Blob/slice()}} with {{end}} normalized according to the following:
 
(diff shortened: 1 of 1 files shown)


index.bs
```diff
@@ -504,7 +500,7 @@ It must act as follows:
     <li>Else, let |relativeStart| be <code>min(start, size)</code>.
   </ol>
 
-3. The optional <dfn argument for="Blob/slice(start, end, contentType)" id="dfn-end">end</dfn> parameter
+2. The optional <dfn argument for="Blob/slice(start, end, contentType)" id="dfn-end">end</dfn> parameter
```
inexorabletash, line 503: Since the spec is already inconsistent about use of 1/1/1/... vs. 1/2/3/..., how about converting this to 1/1/1/... while you're here to save work later?

index.bs
```diff
@@ -1187,7 +1183,7 @@ The {{FileReader}} object must be the event target for all events in this specif
 
 When this specification says to <dfn>fire a progress event</dfn> <i>called e</i>
 (for some {{ProgressEvent}} <code>e</code>
-at a given {{FileReader}} <code>reader</code> as the <a>context object</a>),
+at a given {{FileReader}} <code>reader</code> as <a>this</a>),
```
inexorabletash, line 1186: Just because "as this" reads poorly, how about just dropping that part of the sentence? I don't think it adds anything.

index.bs
```diff
@@ -952,10 +948,10 @@ and runs the associated step:
...
 The <dfn attribute for=FileReader>result</dfn> attribute's getter,
-when invoked, must return the [=context object=]'s [=FileReader/result=].
+when invoked, must return [=this=]'s [=FileReader/result=].
 
 The <dfn attribute for=FileReader>error</dfn> attribute's getter,
-when invoked, must return the [=context object=]'s [=FileReader/error=].
+when invoked, must return [=this=]'s [=FileReader/error=].
 
 <div algorithm="read operation">
 A {{FileReader}} |fr| has an associated <dfn id=readOperation>read operation</dfn> algorithm,
```
inexorabletash, line 957: WDYT about replacing |fr| references in this algorithm with [=this=] ?

inexorabletash: And same for "A {{Blob}} |blob| ..." above (line 266, which I can't comment on here)

inexorabletash (commented):
Drive-by...

index.bs
domenic: [=this=] only has meaning inside method steps/getter steps/setter steps/constructor steps, so I don't think that works.

index.bs
annevk: I think it's meant to say that when the spec says "fire a progress event" you have to use this as the target? But this could do with a rewrite.

I'm happy to drop "as this" though, but I'm not sure if that helps or hurts. (I'm not signing up for the bigger fix.)

mkruisselbrink: dropping "as this" sounds good to me, agree that this could do with a rewrite but that indeed seems like a separate issue.

mkruisselbrink (approved):

PR #169, opened 2021-04-21
Remove Previous Version from .bs file

mkruisselbrink:
When publishing to WD the correct previous version gets added anyway, and having (an incorrect) previous version mentions in the .bs file actually breaks auto-publishing. 

This should fix `"Retrieved "previous" and "latest" documents, but their contents don't match.` errors from echidna.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/pull/169.html" title="Last updated on Apr 21, 2021, 11:21 PM UTC (4a01382)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/169/dcff0c6...4a01382.html" title="Last updated on Apr 21, 2021, 11:21 PM UTC (4a01382)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -11,7 +11,6 @@ Status: ED
 ED: https://w3c.github.io/FileAPI/
 TR: https://www.w3.org/TR/FileAPI/
 Repository: w3c/FileAPI
-Previous Version: https://www.w3.org/TR/2019/WD-FileAPI-20190531/
 !Tests: <a href=https://github.com/web-platform-tests/wpt/tree/master/FileAPI>web-platform-tests FileAPI/</a> (<a href=https://github.com/web-platform-tests/wpt/labels/FileAPI>ongoing work</a>)
 Abstract: This specification provides an API for representing file objects in web applications,
   as well as programmatically selecting them and accessing their data. This includes:


marcoscaceres:
BS'a docs say

> Previous Version can contain either a URL to a previously-published version (such as versions published on w3.org/TR), or the value from biblio or from biblio my-spec-1.

So maybe "from biblio" should go there?

mkruisselbrink:
> BS'a docs say
> 
> > Previous Version can contain either a URL to a previously-published version (such as versions published on w3.org/TR), or the value from biblio or from biblio my-spec-1.
> 
> So maybe "from biblio" should go there?

That seems to work to get the correct version in the ED version, but with the way spec-prod invokes bikeshed will still result in the previous version being listed twice (since spec-prod also passes the latest previous version via the command line). I guess it should at least let echidna not complain?

mkruisselbrink:
So I guess (without changing spec-prod) we can either have previous version in the ED, but have it twice in the WD. Or alternatively just have previous version in the WD and not in the ED at all. Do EDs typically list previous versions?

marcoscaceres:
> I guess it should at least let echidna not complain?

Might be enough.

>  Do EDs typically list previous versions?

No, not typically. But maybe ok if it's not too confusing 🤷 ?

Cc'ing @sidvishnoi, as he might be interested in this discussion.

sideshowbarker:
Currently the way that works with spec-prod is to completely drop the `Previous Version` part from the source, and let spec-prod add it. For example, see https://github.com/w3c/webappsec-csp/pull/484/files.

So the current patch in this PR is doing the right thing.

> Do EDs typically list previous versions?

I don’t know if most EDs have Previous Version links but regardless it doesn’t seem useful for EDs to have Previous Version links.

The Previous Version links have some usefulness in the TR publications, and so outside of the fact they’re required by pubrules for TR publication, it also makes some sense as far as usability/UX for readers to have them there.

But I don’t think that really holds for EDs.

> So I guess (without changing spec-prod) we can either have previous version in the ED, but have it twice in the WD. Or alternatively just have previous version in the WD and not in the ED at all.

It seems fine to have at the Previous Version info just in the WD but the ED. It doesn’t seem like readers are losing anything by us not having the Previous Version link in the ED.

sideshowbarker (approved):

sideshowbarker:
OK that worked:

* https://www.w3.org/TR/FileAPI/ 
* https://www.w3.org/TR/2021/WD-FileAPI-20210422/

PR #172, opened 2021-05-11
Editorial: Remove any references to "parsable MIME type"

### LINKED ISSUE #170: Issues with MIME types
As part of my work on [defining the `multipart/form-data` parser](https://github.com/andreubotella/multipart-form-data), I noticed a couple things in the File API standard related to MIME types:

- The MIME Sniffing standard's "parsable MIME type" definition is no longer present. It was removed as part of whatwg/mimesniff#36.
- Blob's `type` attribute is defined as an ASCII string, perhaps assuming that every string that would be successfully parsed as a MIME type, as well as every serialization of a MIME type, would be ASCII strings. This is not the case, as per whatwg/mimesniff#141.
- Blob's `type` attribute is also defined as being lowercase. And while the MIME parsing algorithm does ASCII-lowercase the MIME record's type, subtype, and parameter names, it doesn't lowercase parameter values. See whatwg/html#6251 for a case where it matters.

The second point implies that on a response coming from the `fetch` API that for whatever reason happened to have the MIME type `multipart/form-data; boundary=cadena-de-separación` (notice the ó), `response.formData()` might succeed but `response.blob()` would fail, which would seem paradoxical.

The third point implies that if, for whatever reason, there was a response coming from the `fetch` API which contained an actual `multipart/form-data` payload coming from Chromium or WebKit (since they start their boundary strings with `WebKitFormBoundary`), and a developer decoded it as a `Blob`; trying to parse that blob afterwards with `new Request(blob).formData()` would fail, since the boundary string seems to be parsed case-sensitively.

andreubotella:
The term "parsable MIME type" used to be part of the MIME Sniffing standard, but it was removed in whatwg/mimesniff#36. This change replaces its uses with equivalent phrasing that references the "parse a MIME type" algorithm. It also replaces mentions of "ASCII-encoded strings" with the Infra standard's definition of "ASCII string".

Closes #170.

Implementation commitment:

 * [ ] WebKit (https://bugs.webkit.org/show_bug.cgi?id=)
 * [ ] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=)
 * [ ] Gecko (https://bugzilla.mozilla.org/show_bug.cgi?id=)

***
<a href="https://pr-preview.s3.amazonaws.com/andreubotella/FileAPI/pull/172.html" title="Last updated on May 11, 2021, 1:29 PM UTC (3e88726)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/172/a7f2705...andreubotella:3e88726.html" title="Last updated on May 11, 2021, 1:29 PM UTC (3e88726)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -68,9 +68,6 @@ spec: url
 </pre>
 
 <pre class="anchors">
-spec: mimesniff; urlPrefix: https://mimesniff.spec.whatwg.org/
-  type: dfn
-    text: parsable mime type
 spec: ecma-262; urlPrefix: http://tc39.github.io/ecma262/
   type: interface
     text: Array; url: sec-array-constructor
@@ -208,7 +205,7 @@ this is the same time that is conceptually "<code>0</code>" in ECMA-262 [[ECMA-2
 A {{Blob}} object refers to a <a>byte</a> sequence,
 and has a {{Blob/size}} attribute which is the total number of bytes in the byte sequence,
 and a {{Blob/type}} attribute,
-which is an ASCII-encoded string in lower case representing the media type of the <a>byte</a> sequence.
+which is an [=ASCII string=] in lower case representing the media type of the <a>byte</a> sequence.
 
 Each {{Blob}} must have an internal <dfn id="snapshot-state" for=Blob>snapshot state</dfn>,
 which must be initially set to the state of the underlying storage,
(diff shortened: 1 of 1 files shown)


annevk:
You pointed out that MIME types aren't necessarily ASCII so presumably we shouldn't say that here either? Or is it still incorrectly stripping parameters?

PR #183, opened 2022-11-10
editorial: make Blob.slice a separate algorithm

dlrobertson:
Split out Blob.slice into its own algorithm that other implementations may use and call this algorithm from Blob.slice.

Related to: https://github.com/whatwg/fetch/issues/1070

***
<a href="https://pr-preview.s3.amazonaws.com/dlrobertson/FileAPI/pull/183.html" title="Last updated on Nov 13, 2022, 8:26 PM UTC (4b840e6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/183/57c12ca...dlrobertson:4b840e6.html" title="Last updated on Nov 13, 2022, 8:26 PM UTC (4b840e6)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -193,6 +193,76 @@ The term <dfn id="UnixEpoch">Unix Epoch</dfn> is used in this specification to r
 (or 1970-01-01T00:00:00Z ISO 8601);
 this is the same time that is conceptually "<code>0</code>" in ECMA-262 [[ECMA-262]].
 
+<div algorithm>
+The <dfn export>slice blob</dfn> algorithm given a {{Blob}} <var>blob</var>,
+an optional parameter <var>start</var>, an optional parameter <var>end</var>,
+and an optional parameter <var>contentType</var> is used to refer to the following
+steps and returns a new {{Blob}} containing the bytes ranging from the <var>start</var> parameter
+up to but not including the <var>end</var> parameter. It must act as follows:
+
+1. Let <var>originalSize</var> be <var>blob</var>'s {{Blob/size}}.
+
+1. The optional <var>start</var> parameter is a value for the start point of a <a>slice blob</a>
+   call, and must be treated as a byte-order position, with the zeroth position representing the
+   first byte. User agents must normalize <var>start</var> according to the following:
+
+  <ol type="a">
+   <li>If the optional <var>start</var> parameter is not used as a parameter when making this call,
+   let <var>relativeStart</var> be 0.
+
+   <li>If <var>start</var> is negative, let <var>relativeStart</var> be
+   <code>max((<var>originalSize</var> + <var>start</var>), 0)</code>.
+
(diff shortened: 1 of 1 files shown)


dlrobertson:
Just a first pass at the moment. Still likely have a good bit of cleanup to do. Originally I wanted a normalize range and a slice bytes algorithm, but after further reading I think its probably easier to just bundle all of it in one algorithm (especially since slice bytes is really only one line in the algorithm).

dlrobertson:
I don't think this is substantive, but if this is considered to be substantive, I can create a w3c account.

index.bs
```diff
@@ -193,6 +193,75 @@ The term <dfn id="UnixEpoch">Unix Epoch</dfn> is used in this specification to r
...
+        substeps.
+      2. Convert every character in <var>relativeContentType</var> to [=ASCII lowercase=].
+  </ol>
+
+1. Let <var>span</var> be <code>max((<var>relativeEnd</var> - <var>relativeStart</var>), 0)</code>.
+
+1. Return a new {{Blob}} object <var>S</var> with the following characteristics:
+
+  <ol type="a">
+    <li><var>S</var> refers to <var>span</var> consecutive <a>byte</a>s from <var>blob</var>'s,
```
mkruisselbrink, line 254:
```suggestion
    <li><var>S</var> refers to <var>span</var> consecutive <a>byte</a>s from <var>blob</var>'s
```

index.bs
```diff
@@ -193,6 +193,75 @@ The term <dfn id="UnixEpoch">Unix Epoch</dfn> is used in this specification to r
 (or 1970-01-01T00:00:00Z ISO 8601);
 this is the same time that is conceptually "<code>0</code>" in ECMA-262 [[ECMA-262]].
 
+<div algorithm>
+The <dfn id="slice-blob">slice blob</dfn> algorithm given a {{Blob}} <var>blob</var>,
```
mkruisselbrink, line 197: You probably want this to be exported so you can reference it from fetch? And this should probably be an abstract-op, and I don't think it needs an explicit ID.

```suggestion
The <dfn abstract-op export>slice blob</dfn> algorithm given a {{Blob}} <var>blob</var>,
```

mkruisselbrink (commented):
Generally looks good, just two minor points. I agree that just moving stuff around like this is probably not substantive. I really need to get around to rewriting splice more formally at some point, but that is unrelated to this.

index.bs
annevk: It should not be an abstract-op (that's mainly JS stuff, all our algorithms are the default dfn most of the time), but you're right that relying on the generated ID is fine.

dlrobertson (commented):
Thanks for the comments! Fixed the remaining build errors and made the suggested changes.

mkruisselbrink (approved):

PR #184, opened 2022-11-26
Editorial: slice blob should use nullable args

dlrobertson:
Use nullable arguments instead of optional arguments for the exported slice blob algorithm to make it easier to be called from other specs.

Related to: https://github.com/whatwg/fetch/pull/1520

***
<a href="https://pr-preview.s3.amazonaws.com/dlrobertson/FileAPI/pull/184.html" title="Last updated on Nov 28, 2022, 3:16 PM UTC (21f0d56)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/184/3ee8d23...dlrobertson:21f0d56.html" title="Last updated on Nov 28, 2022, 3:16 PM UTC (21f0d56)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -194,21 +194,19 @@ The term <dfn id="UnixEpoch">Unix Epoch</dfn> is used in this specification to r
 this is the same time that is conceptually "<code>0</code>" in ECMA-262 [[ECMA-262]].
 
 <div algorithm>
-The <dfn export>slice blob</dfn> algorithm given a {{Blob}} <var>blob</var>,
-an optional parameter <var>start</var>, an optional parameter <var>end</var>,
-and an optional parameter <var>contentType</var> is used to refer to the following
+The <dfn export>slice blob</dfn> algorithm given a {{Blob}} <var>blob</var>, <var>start</var>,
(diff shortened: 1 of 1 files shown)


dlrobertson:
Like https://github.com/w3c/FileAPI/pull/183, I do not believe this to be substantive.

index.bs
```diff
@@ -558,8 +555,20 @@ returns a new {{Blob}} object with bytes ranging from the optional <var>start</v
 up to but not including the optional <var>end</var> parameter, and with a {{Blob/type}} attribute
 that is the value of the optional <var>contentType</var> parameter. It must act as follows:
 
-1. Return the result of <a>slice blob</a> given <a>this</a>, <var>start</var>, <var>end</var>,
-   and <var>contentType</var>.
+1. Let <var>sliceStart</var> be null.
```
annevk, line 558: Nit: you could initialize all three to null here at once to make it slightly shorter.

annevk (approved):

index.bs
dlrobertson: Good point!

mkruisselbrink (approved):

PR #191, opened 2023-02-03
Update the exported text for resolve

### LINKED ISSUE #190: Don't export "resolve"
Any spec author who wants to reference https://webidl.spec.whatwg.org/#resolve through `[=resolve=]` will run into conflicts with this specification, so we should either avoid exporting or specialize the name of this export.

johannhof:
Closes #190 

I'll update https://url.spec.whatwg.org/#url-parsing as soon as this PR merges.

***
<a href="https://pr-preview.s3.amazonaws.com/johannhof/FileAPI/pull/191.html" title="Last updated on Feb 6, 2023, 12:22 PM UTC (e0f1727)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/191/d134dae...johannhof:e0f1727.html" title="Last updated on Feb 6, 2023, 12:22 PM UTC (e0f1727)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1558,7 +1558,7 @@ run the following steps:
 ## Dereferencing Model for blob URLs ## {#requestResponseModel}
 
 <div algorithm="resolveURL">
-To <dfn export>resolve a blob URL</dfn> given a |url| (a [=URL=]), run the following steps:
+To <dfn export id=blob-url-resolve>resolve a blob URL</dfn> given a [=URL=] |url|:
 
 1. [=Assert=]: |url|'s [=url/scheme=] is "`blob`".
 1. Let |store| be the user agent's [=blob URL store=].
--- a/index.bs
+++ b/index.bs
@@ -1558,7 +1558,7 @@ run the following steps:
 ## Dereferencing Model for blob URLs ## {#requestResponseModel}
 
 <div algorithm="resolveURL">
-To <dfn export for="blob URL" lt="resolve">resolve a blob URL</dfn> given a |url| (a [=URL=]), run the following steps:
+To <dfn export for="blob URL">resolve a blob URL</dfn> given a |url| (a [=URL=]), run the following steps:
 
 1. [=Assert=]: |url|'s [=url/scheme=] is "`blob`".
 1. Let |store| be the user agent's [=blob URL store=].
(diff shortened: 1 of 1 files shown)


saschanaz (approved):

index.bs
```diff
@@ -1558,7 +1558,7 @@ run the following steps:
 ## Dereferencing Model for blob URLs ## {#requestResponseModel}
 
 <div algorithm="resolveURL">
-To <dfn export for="blob URL" lt="resolve">resolve a blob URL</dfn> given a |url| (a [=URL=]), run the following steps:
+To <dfn export for="blob URL">resolve a blob URL</dfn> given a |url| (a [=URL=]), run the following steps:
```
mkruisselbrink, line 1561: With `resolve a blob URL` as text, I'd get rid of the `for` part, as that is redundant now.

mkruisselbrink (commented):
This seems like kind of a unnecessary change, but I suppose it doesn't hurt...

annevk:
I agree with @mkruisselbrink that this isn't needed. This is what `for` is for. We'd have to change a whole lot of terms across a whole lot of specifications if we want by this rationale.

johannhof:
> I agree with @mkruisselbrink that this isn't needed. This is what `for` is for. We'd have to change a whole lot of terms across a whole lot of specifications if we want by this rationale.

Sure, I agree as well, but https://github.com/w3c/FileAPI/issues/190#issuecomment-1418896202

> Edit: having said that, in this specific case the change does kinda make sense as the concept doesn't really belong to the blob URL concept.

Ok, well, I'll remove the `for="blob URL"` then :)

index.bs
```diff
@@ -1558,7 +1558,7 @@ run the following steps:
 ## Dereferencing Model for blob URLs ## {#requestResponseModel}
 
 <div algorithm="resolveURL">
-To <dfn export for="blob URL" lt="resolve">resolve a blob URL</dfn> given a |url| (a [=URL=]), run the following steps:
+To <dfn export>resolve a blob URL</dfn> given a |url| (a [=URL=]), run the following steps:
```
annevk, line 1561: Let's preserve the ID and modernize a bit while we're here.
```suggestion
To <dfn export id=blob-url-resolve>resolve a blob URL</dfn> given a [=URL=] |url|:
```

Issue #153, opened 2020-05-05
Blob URL store partitioning


annevk:
https://privacycg.github.io/storage-partitioning/ has some general background here and https://trac.torproject.org/projects/tor/ticket/15502 is much more specific.

@bakulf was thinking that we could restrict blob URL lookup to the agent cluster (in addition to origin, that is). The one tweak I would suggest to that is that navigating a top-level browsing context (including a noopener one) to a blob URL still ought to work.

Concretely, this would mean that if you have `https://example.com/` open twice, in separate browsing context groups, any blob URLs they mint cannot be used by the other.

The one gotcha with the tweak I suggested is that the other could observe existence through a popup then. Now that's an attack that's unlikely to yield anything useful in practice, but we could break that too by forcing noopener or a version of COOP that never matches (and thus always creates a new browsing context group).

We suspect this to be web-compatible and are happy to try it out in Firefox.

cc @mkruisselbrink @hober @SubhamoyS

mkruisselbrink:
Limiting it like that doesn't seem too crazy to me, although I'm not entirely sure I understand the attack/threat model that defends against.

A blob URL can only be resolved when the page that created it is still alive, and when the page trying to fetch it is same origin with the original page. If that is the case, you might as well just BroadcastChannel to talk to the original page, rather than jumping through hoops with blob URLs?

Of course if storage partitioning blocks BroadcastChannel because of different top-level URLs things would be different, but in that case wouldn't it make more sense to partition blob URLs the same way as other communication mechanisms, rather than having them be partitioned even more?

mkruisselbrink:
Having said all that, for MediaSource blob URLs chrome already limits them to their agent cluster. And scoping all blob URLs to their agent cluster certainly would make the implementation a whole lot simpler (well, the top-level navigation case would still be tricky). So as an implementer agent cluster seems reasonable, as a spec editor I question if that is the right scoping level.

annevk:
The keying for storage seems like it will end up being dynamic (i.e., storage access). I think if we can avoid things being dynamically keyed that's preferable.

smaug----:
It would be a bit odd inconsistency if sharing objects worked, but sharing blob URLs didn't.

mkruisselbrink:
That doesn't seem that different from today? Blob objects can be shared across origins, while blob URLs can only be resolved by same origin originators. As such there are already plenty of cases where sharing a blob URL doesn't work but sharing a blob, and having the receiver create a blob URL from it does work.

[...]

PR #201, opened 2024-10-25
Partition Blob URL revocation by Storage Key

recvfrom:
Part of the changes discussed in https://github.com/w3c/FileAPI/issues/153#issuecomment-2330085478

This updates URL.revokeObjectURL to not allow revoking a Blob URL except from contexts with the same Storage Key as the one in which the Blob URL was created. A corresponding PR will update the Fetch spec to incorporate similar Storage Key checks into Blob URL fetches.

I considered incorporating the Storage Key checks into the "resolve a blob URL" algorithm instead, but it seemed that this would require an environment settings object to be available as part of https://url.spec.whatwg.org/#url-parsing, and I'm not sure whether that is the case / a change we want.

 * [x] Modified Web platform tests (link to pull request)
 - https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/web_tests/external/wpt/FileAPI/BlobURL/cross-partition.tentative.https.html

Implementation commitment:

 * [x] WebKit - Already implemented partitioning by top-level origin, considering partitioning by top-level site - https://github.com/w3c/FileAPI/issues/153#issuecomment-2332086739
 * [x] Chromium (https://bugs.chromium.org/p/chromium/issues/detail?id=40057646)
 * [x] Gecko - Already implemented - https://github.com/w3c/FileAPI/issues/153#issuecomment-2332288047

***
<a href="https://pr-preview.s3.amazonaws.com/recvfrom/FileAPI/pull/201.html" title="Last updated on Dec 4, 2024, 5:29 AM UTC (5a9047c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/201/f6247a2...recvfrom:5a9047c.html" title="Last updated on Dec 4, 2024, 5:29 AM UTC (5a9047c)">Diff</a>

### DIFF
--- a/index.bs
+++ b/index.bs
@@ -1512,14 +1512,24 @@ A [=blob URL store=] is a [=map=]
 where [=map/keys=] are [=valid URL strings=]
 and [=map/values=] are [=blob URL Entries=].
 
-A <dfn export>blob URL entry</dfn> consists of
-an <dfn export for="blob URL entry">object</dfn> (of type {{Blob}} or {{MediaSource}}),
-and an <dfn export for="blob URL entry">environment</dfn> (an [=environment settings object=]).
-
 [=map/Keys=] in the [=blob URL store=] (also known as <dfn lt="blob URL|object URL" export>blob URLs</dfn>)
 are [=valid URL strings=] that when [=URL parser|parsed=]
 result in a [=/URL=] with a [=url/scheme=] equal to "`blob`",
 an [=empty host=], and a [=url/path=] consisting of one element itself also a [=valid URL string=].
+
+A <dfn export>blob URL entry</dfn> is used to store an object of type {{Blob}} or {{MediaSource}}.
+
+A [=blob URL entry=] has an <dfn export for="blob URL entry">environment</dfn> (an [=environment settings object=]).
+
+<div algorithm="obtainBlobObject">
+To <dfn export id=blob-url-obtain-object>obtain a blob object</dfn> given a [=blob URL entry=] |blobUrlEntry|, an [=environment=] |environment|, and an optional boolean |isNavigation| (default false):
+
+1. Let |isAuthorized| be true.
+1. If |isNavigation| is false, let |isAuthorized| be the result of [=checking for same-partition blob URL usage=] with |blobUrlEntry| and |environment|.
+1. If |isAuthorized| is false, then return failure.
+1. Return |blobUrlEntry|'s object.
+
+</div>
 
 <div algorithm="createBlobURL">
 To <dfn id="unicodeBlobURL" lt="generate a new blob URL|generating a new blob URL">
@@ -1595,6 +1605,22 @@ as the serialization of the origin of the environment that created the blob URL,
 but for opaque origins the origin itself might be distinct. This difference isn't
 observable though, since a revoked blob URL can't be resolved/fetched anymore anyway.
 
+### Access restrictions on blob URLs ### {#partitioningOfBlobUrls}
+
+<a>Blob URLs</a> can only be fetched from environments where the [=storage key=] matches that of
+the environment where the <a>blob URL</a> was created. <a>blob URL</a> navigations are not subject
+to this restriction.
+
+<div algorithm="checkForSamePartitionBlobUrlUsage">
+To <dfn export id=blob-url-partition-check>check for same-partition blob URL usage</dfn> given a [=blob URL entry=] |blobUrlEntry| and an [=environment=] |environment|:
+
+1. Let |blobStorageKey| be the result of [=obtaining a storage key for non-storage purposes=] with |blobUrlEntry|'s [=blob URL entry/environment=].
+1. Let |environmentStorageKey| be the result of [=obtaining a storage key for non-storage purposes=] with |environment|.
+1. If |blobStorageKey| is not [=storage key/equal=] to |environmentStorageKey|, then return false.
+1. Otherwise, return true.
+
+</div>
+
 <h4 id="lifeTime" dfn for="blob url" lt="lifetime|lifetime stipulation" export>Lifetime of blob URLs</h4>
 
 This specification extends the [=unloading document cleanup steps=] with the following steps:
@@ -1633,12 +1659,12 @@ The <dfn method for=URL id="dfn-revokeObjectURL">revokeObjectURL(|url|)</dfn> st
 1. Let |url record| be the result of [=URL parser|parsing=] |url|.
 1. If |url record|'s [=url/scheme=] is not "`blob`", return.
 1. Let |entry| be |url record|'s [=blob URL entry=].
-1. Let |blobStorageKey| be the result of [=obtaining a storage key for non-storage purposes=] with |entry|'s [=blob URL entry/environment=].
-1. Let |currentStorageKey| be the result of [=obtaining a storage key for non-storage purposes=] with the [=current settings object=].
-1. If |blobStorageKey| is not [=storage key/equal=] to |currentStorageKey|, return.
+1. If |entry| is null, return.
+1. Let |isAuthorized| be the result of [=checking for same-partition blob URL usage=] with |entry| and the [=current settings object=].
+1. If |isAuthorized| is false, return.
 1. [=Remove an entry from the Blob URL Store=] for |url|.
 
-Note: This means that rather than throwing some kind of error, attempting to revoke a URL that isn't registered or that was registered from an environment with a different storage key will silently fail.
+Note: This means that rather than throwing some kind of error, attempting to revoke a URL that isn't registered or that was registered from an environment in a different storage partition will silently fail.
 User agents might display a message on the error console if this happens.
 
 Note: Attempts to dereference |url| after it has been revoked will result in a [=network error=].
--- a/index.bs
+++ b/index.bs
@@ -1512,12 +1512,12 @@ A [=blob URL store=] is a [=map=]
 where [=map/keys=] are [=valid URL strings=]
 and [=map/values=] are [=blob URL Entries=].
 
-A <dfn export>blob URL entry</dfn> consists of an <i>object</i> (of type {{Blob}} or
-{{MediaSource}}), and an <dfn export for="blob URL entry">environment</dfn> (an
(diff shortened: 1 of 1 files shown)


miketaylr:
Mind taking a look @mkruisselbrink?

index.bs
```diff
@@ -1630,11 +1630,11 @@ return the result of [=adding an entry to the blob URL store=] for |obj|.
 <div algorithm="revokeObjectURL">
 The <dfn method for=URL id="dfn-revokeObjectURL">revokeObjectURL(|url|)</dfn> static method must run these steps:
 
-1. Let |url record| be the result of [=URL parser|parsing=] |url|.
-1. If |url record|'s [=url/scheme=] is not "`blob`", return.
-1. Let |origin| be the [=url/origin=] of |url record|.
-1. Let |settings| be the [=current settings object=].
-1. If |origin| is not [=same origin=] with |settings|'s [=environment settings object/origin=], return.
+1. Let |entry| be the result of [=resolving the blob URL=] |url|.
```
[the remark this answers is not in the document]
mkruisselbrink: Also it seems like you don't need to explicitly invoke resolve the blob URL, rather keep the old algorithm and reference `|url record|'s blob URL entry` would get you the entry you need already?

index.bs
```diff
@@ -1630,11 +1630,11 @@ return the result of [=adding an entry to the blob URL store=] for |obj|.
...
+1. If |entry| is failure, return.
+1. Let |blobStorageKey| be the result of [=obtaining a storage key for non-storage purposes=] with |entry|'s [=environment settings object=].
```
mkruisselbrink, line 1635:
```suggestion
1. Let |blobStorageKey| be the result of [=obtaining a storage key for non-storage purposes=] with |entry|'s [=blob URL entry/environment=].
```

index.bs
```diff
@@ -1630,11 +1630,11 @@ return the result of [=adding an entry to the blob URL store=] for |obj|.
...
+1. Let |currentStorageKey| be the result of [=obtaining a storage key for non-storage purposes=] with the [=current settings object=].
+1. If |blobStorageKey| is not [=storage key/equal=] to |currentStorageKey|, return.
 1. [=Remove an entry from the Blob URL Store=] for |url|.
 
 Note: This means that rather than throwing some kind of error, attempting to revoke a URL that isn't registered will silently fail.
```
mkruisselbrink, line 1640: Might make sense to mention that it's not just URLs that aren't registered that would silently fail, something like (although not sure that's the best way to phrase it):

```suggestion
Note: This means that rather than throwing some kind of error, attempting to revoke a URL that isn't registered, or is registered from a different storage key, will silently fail.
```

index.bs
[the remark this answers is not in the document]
recvfrom: yikes I was way off on my initial attempt of this 😅 sorry about that. I think I've addressed the issues you've pointed out. Thanks for the suggestions!

index.bs
recvfrom: Good catch, I updated this (although made it a bit more verbose). WDYT?

miketaylr:
I guess we need to update spec-prod (or wait for the update to be propagated?):

```
  $ bikeshed update
    Bikeshed now requires Python 3.9; you are on 3.8.10.
        If you're seeing this message in your CI run, you are
        likely specifying an old OS; try `ubuntu-latest`.
        If you're seeing this on the command line, see the docs
        for instructions:
        https://speced.github.io/bikeshed/#installing
  Command `bikeshed update` failed with exit code: 1.
```

https://github.com/w3c/spec-prod/pull/193

mkruisselbrink:
> I guess we need to update spec-prod (or wait for the update to be propagated?):
> 
> ```
>   $ bikeshed update
>     Bikeshed now requires Python 3.9; you are on 3.8.10.
>         If you're seeing this message in your CI run, you are
>         likely specifying an old OS; try `ubuntu-latest`.
>         If you're seeing this on the command line, see the docs
>         for instructions:
>         https://speced.github.io/bikeshed/#installing
>   Command `bikeshed update` failed with exit code: 1.
> ```
> 
> [w3c/spec-prod#193](https://github.com/w3c/spec-prod/pull/193)

I think https://github.com/w3c/FileAPI/pull/204 will fix that.

recvfrom:
Per the feedback here [1], I've updated this PR to also change how the blob URL entry's object member is exposed so that the partitioning checks can be fully contained in the FileAPI. WDYT?

[1] https://github.com/whatwg/fetch/pull/1783#issuecomment-2493209801

index.bs
```diff
@@ -1512,15 +1512,25 @@ A [=blob URL store=] is a [=map=]
...
-A <dfn export>blob URL entry</dfn> consists of
-an <dfn export for="blob URL entry">object</dfn> (of type {{Blob}} or {{MediaSource}}),
-and an <dfn export for="blob URL entry">environment</dfn> (an [=environment settings object=]).
-
 [=map/Keys=] in the [=blob URL store=] (also known as <dfn lt="blob URL|object URL" export>blob URLs</dfn>)
 are [=valid URL strings=] that when [=URL parser|parsed=]
 result in a [=/URL=] with a [=url/scheme=] equal to "`blob`",
 an [=empty host=], and a [=url/path=] consisting of one element itself also a [=valid URL string=].
 
+A <dfn export>blob URL entry</dfn> is used to store an object of type {{Blob}} or {{MediaSource}}.
```
annevk, line 1520: Looking at this. I would continue to give a blob URL entry an object field, but just no longer export it. And then explain in a note that specifications have to use "obtain a blob object" to get hold of the object.

index.bs
```diff
@@ -1512,15 +1512,25 @@ A [=blob URL store=] is a [=map=]
...
+
+A [=blob URL entry=] has an <dfn export for="blob URL entry">environment</dfn> (an [=environment settings object=]).
+
+<div algorithm="obtainBlobObject">
+To <dfn export id=blob-url-obtain-object>obtain a blob object</dfn> given a [=blob URL entry=] |blobUrlEntry|, an [=environment=] |environment|, and an optional boolean |isNavigation| (default false):
+
+1. Let |isAuthorized| be true.
+1. If |isNavigation| is false, let |isAuthorized| be the result of [=checking for same-partition blob URL usage=] with |blobUrlEntry| and |environment|.
```
annevk, line 1528: If you manipulate an initialized variable, you need to use "set ... to". See the Infra standard for more complete guidance.

index.bs
```diff
@@ -1632,12 +1658,13 @@ The <dfn method for=URL id="dfn-revokeObjectURL">revokeObjectURL(|url|)</dfn> st
 
 1. Let |url record| be the result of [=URL parser|parsing=] |url|.
```
annevk, line 1659: Maybe change this to urlRecord while here? Having these as separate words is confusing with the url variable.

index.bs
```diff
@@ -1632,12 +1658,13 @@ The <dfn method for=URL id="dfn-revokeObjectURL">revokeObjectURL(|url|)</dfn> st
...
 1. If |url record|'s [=url/scheme=] is not "`blob`", return.
-1. Let |origin| be the [=url/origin=] of |url record|.
-1. Let |settings| be the [=current settings object=].
-1. If |origin| is not [=same origin=] with |settings|'s [=environment settings object/origin=], return.
+1. Let |entry| be |url record|'s [=blob URL entry=].
+1. If |entry| is null, return.
```
annevk, line 1662: then return*

index.bs
```diff
@@ -1632,12 +1658,13 @@ The <dfn method for=URL id="dfn-revokeObjectURL">revokeObjectURL(|url|)</dfn> st
...
+1. Let |isAuthorized| be the result of [=checking for same-partition blob URL usage=] with |entry| and the [=current settings object=].
+1. If |isAuthorized| is false, return.
```
annevk, line 1664: then return*

index.bs
recvfrom: Thanks for the feedback, I hadn't thought to just not export the field. I think I've implemented this as you suggested but let me know if you have any other thoughts on this. Thanks!

index.bs
```diff
@@ -1595,6 +1609,22 @@ as the serialization of the origin of the environment that created the blob URL,
...
+the environment where the <a>blob URL</a> was created. <a>blob URL</a> navigations are not subject
+to this restriction.
+
+<div algorithm="checkForSamePartitionBlobUrlUsage">
+To <dfn export id=blob-url-partition-check>check for same-partition blob URL usage</dfn> given a [=blob URL entry=] |blobUrlEntry| and an [=environment settings object=] |environment|:
+
+1. Let |blobStorageKey| be the result of [=obtaining a storage key for non-storage purposes=] with |blobUrlEntry|'s [=blob URL entry/environment=].
+1. Let |environmentStorageKey| be the result of [=obtaining a storage key for non-storage purposes=] with |environment|.
+1. If |blobStorageKey| is not [=storage key/equal=] to |environmentStorageKey|, then return false.
+1. Otherwise, return true.
```
annevk, line 1624: No need for Otherwise when you return early.
```suggestion
1. Return true.
```

annevk (commented):
This looks good to me modulo nits. Would be good if @mkruisselbrink had another look as well.

index.bs
```diff
@@ -1595,6 +1609,22 @@ as the serialization of the origin of the environment that created the blob URL,
 but for opaque origins the origin itself might be distinct. This difference isn't
 observable though, since a revoked blob URL can't be resolved/fetched anymore anyway.
 
+### Access restrictions on blob URLs ### {#partitioningOfBlobUrls}
+
+<a>Blob URLs</a> can only be fetched from environments where the [=storage key=] matches that of
+the environment where the <a>blob URL</a> was created. <a>blob URL</a> navigations are not subject
```
mkruisselbrink, line 1615:
```suggestion
the environment where the <a>blob URL</a> was created. <a>Blob URL</a> navigations are not subject
```

(capitalize start of sentence)

mkruisselbrink (commented):
looks good to me as well, modulo Anne's nits

index.bs
recvfrom: Good catch! Updated

index.bs
recvfrom: Good catch, thanks

mkruisselbrink (approved):

PR #222, opened 2026-07-17
Fold File and Directory Entries API into File API

jcayzac:
Incorporates the File-API-bound content from the WICG Entries API (https://wicg.github.io/entries-api/) into this specification:

- Path and entry concepts (names, paths, files, directories, entries) and the resolve/evaluate-a-path algorithms, added to the Terminology and Algorithms section.
- `File.webkitRelativePath` was merged into the `File` IDL.
- New "Files and Directories" section with IDL for `FileSystemEntry`, `FileSystemDirectoryEntry`, `FileSystemDirectoryReader`, `FileSystemFileEntry` and `FileSystem`, along with all associated callbacks, dictionaries, algorithms, and examples.

* [ ] HTML-bound content (`HTMLInputElement.webkitdirectory` / `webkitEntries` and `DataTransferItem.webkitGetAsEntry()`) remains to be added to the HTML standard.
* [ ] Technical issues are not numbered yet —if the edits are in line with what the WG wants I'll create those and add them.

Implementation commitments are "not applicable" for now since this reflects the current state of implementations.

@marcoscaceres @mkruisselbrink is this directionally correct / what you had in mind?

***
<a href="https://pr-preview.s3.amazonaws.com/jcayzac/FileAPI/pull/222.html" title="Last updated on Jul 17, 2026, 2:38 AM UTC (0e69ab1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/FileAPI/222/cd1d1da...jcayzac:0e69ab1.html" title="Last updated on Jul 17, 2026, 2:38 AM UTC (0e69ab1)">Diff</a>


index.bs
```diff
@@ -75,6 +78,9 @@ spec: ecma-262; urlPrefix: http://tc39.github.io/ecma262/
   type: interface
     text: Array; url: sec-array-constructor
     text: Date; url: sec-date-constructor
+  type: dfn
```
marcoscaceres, line 81: Nit: it's weird that we need this here... bikeshed should have found it on its own, but I might be wrong. Might be worth a quick check with removing it.

index.bs
```diff
@@ -260,6 +266,179 @@ up to but not including the <var>end</var> parameter. It must act as follows:
 
 </div>
 
+## Names and paths ## {#names-paths}
+
+A <dfn export>name</dfn> is a string which:
```
marcoscaceres, line 271: Maybe... just exporting "name" is too generic: 
```suggestion
A <dfn export local-lt="name">file name</dfn> is a [=string=] which:
```

index.bs
```diff
@@ -260,6 +266,179 @@ up to but not including the <var>end</var> parameter. It must act as follows:
...
+
+  * does not contain '/' (U+002F SOLIDUS)
+  * does not contain NUL (U+0000)
+  * does not contain '\' (U+005C REVERSE SOLIDUS)
+  * is not '.' (U+002E FULL STOP)
+  * is not '..' (U+002E FULL STOP, U+002E FULL STOP)
+
+A <dfn export>path segment</dfn> is a [=/name=], '.' (U+002E FULL STOP) or
```
marcoscaceres, line 279: @annevk, could use your opinion here... should we just do this kind of export, or do you think it might be better to `a path's <dfn for=path export>segment</dfn>`? 

Same with the ones below... 

I'm also worried about collisions with https://url.spec.whatwg.org/#relative-url-string and friends, so want to make sure this is ok.

index.bs
```diff
@@ -260,6 +266,179 @@ up to but not including the <var>end</var> parameter. It must act as follows:
...
+(U+002F SOLIDUS).
+
+A <dfn export>path</dfn> is either a [=relative path=] or an [=absolute
+path=].
+
+A <dfn export>valid path</dfn> is a {{USVString}} which is a [=path=].
+
+## Files and directories ## {#files-dirs}
+
+A <dfn export id=file-concept>file</dfn> consists of binary data and a <dfn
```
marcoscaceres, line 297: How come this one needs an id and the others don't?

index.bs
```diff
@@ -260,6 +266,179 @@ up to but not including the <var>end</var> parameter. It must act as follows:
...
+a distinct non-empty [=/name=].
+
+A <dfn export>root directory</dfn> is a [=/directory=] that is not a member of
+a [=/directory=]. A [=root directory=]'s [=/name=] is empty.
+
+The <dfn export>parent</dfn> of a [=file=] or [=/directory=] is the
+[=/directory=] it is a member of. A [=root directory=] has no
+[=parent=].
+
+Issue: Should [=/directory=] be defined as a special type of [=file=]
```
marcoscaceres, line 312: Do we have a bug filed for this we could point to here?

index.bs
```diff
@@ -260,6 +266,179 @@ up to but not including the <var>end</var> parameter. It must act as follows:
...
+
+A <dfn export>file system</dfn> consists of a <dfn for="file system">name</dfn>
+and a <dfn for="file system">root</dfn> which is an associated [=root
+directory=]. The [=file system/name=] of a [=/file system=] is a
+{{USVString}} which is implementation defined but is unique to the
+[=/file system=]. A [=root directory=] is associated with exactly one
+[=/file system=].
+
+note:
+  Implementations could produce a [=file system/name=] by generating a
```
marcoscaceres, line 329:
```suggestion
  Implementations could produce a file system [=file system/name=] by generating a
```

index.bs
```diff
@@ -704,6 +883,7 @@ interface File : Blob {
               optional FilePropertyBag options = {});
   readonly attribute DOMString name;
   readonly attribute long long lastModified;
+  readonly attribute USVString webkitRelativePath;
```
marcoscaceres, line 886: Argh… do we really want to standardize this here? Should we send this to the WHATWG Compat spec instead?

index.bs
jcayzac: @marcoscaceres not yet. Should I move issues currently in the entries-api repo over here, by the way, or should we keep existing ones there and point to them?

index.bs
jcayzac: @marcoscaceres That's fine by me. I was told merging as-is was higher priority than unprefixing and solving existing issues but I can start doing that now.

So this would need to be replaced with an unprefixed `relativePath` for the purpose of this spec, right? And to get commitment from browsers + have the old prefixed version in Compat.

annevk: I think @jcayzac did the right thing.

The Compat specification is meant to be temporary. Prefixed features that end up widely deployed should just be standardized in the relevant specification, not separately.

We also don't want to introduce new features while making a big move like this.

index.bs
annevk: It might be better to not export this for now, unless we really need it to be exported.

index.bs
marcoscaceres: Ah, sorry, I had this backwards then. I thought it was the other way around.

jcayzac: @annevk @marcoscaceres Thank you. However I think I'm still unclear on what should be the exact course of action.

"Prefixed features that end up widely deployed should just be standardized in the relevant specification" suggests I should unprefix it already since it's implemented everywhere? Or you meant standardize the prefixed version? The last sentence reads like you'd rather not unprefix now.

Sorry progress on this has been delayed a bit due to summer vacations.

annevk: Correct, we should standardize it including prefix as that is what is implemented everywhere. At a future point we can consider standardizing a separate unprefixed version. (In prior discussions that also came with some potential API refinements so it really needs to be separate from this effort.)

